PR #13, opened 2021-07-30
Brief install instructions

rob-luke:
It may be too early to be encouraging users to use this, so feel free to reject.

But this had the ulterior motive of checking with you that I am using the installed software in the way you intended?

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ You do not need to manually install Anaconda before running the installation.
 
 ### MacOS
 
-Move to your home directory and activate the environment:
+Open a terminal and activate the environment:
 
 ```bash
 source ~/mne-python/bin/activate
--- a/README.md
+++ b/README.md
@@ -2,16 +2,18 @@
 
 Installers for MNE-Python for macOS, Windows, and Linux.
 
-⚠️ Nothing to see here for now, please move along and check back again in a few
-days. ⚠️
-
-
 
 ## Installation
 
-To install MNE-Python simply double click the installer.
-You can customise some parameters, but the default values should suit most users.
-Once completed, a conda evironment should be installed in your home directory.
+To install MNE-Python simply:
+* double click the installer on MacOS or Windows, or
+* run the shell script using `sh MNE-Python-0.23.0_installer-2021.1-Linux.sh` on *nix.
+
+You can customize some parameters, but the default values should suit most users.
+By default, a conda evironment will be installed in your home directory,
+however, you can change the install location using the installer options.
+
+You do not need to manually install Anaconda before running the installation.
 
 
 ## Usage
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,3 +4,26 @@ Installers for MNE-Python for macOS, Windows, and Linux.
 
 ⚠️ Nothing to see here for now, please move along and check back again in a few
 days. ⚠️
```
hoechenberger, line 6: We can actually drop this when merging this PR :)

README.md
```diff
@@ -4,3 +4,26 @@ Installers for MNE-Python for macOS, Windows, and Linux.
...
+
+
+
+## Installation
+
+To install MNE-Python simply double click the installer.
```
hoechenberger, line 12: This is only the case for Windows and macOS. On Linux, users will have to run the `.sh` file in the shell via `sh xxx.sh` or `bash xxx.sh`

README.md
```diff
@@ -4,3 +4,26 @@ Installers for MNE-Python for macOS, Windows, and Linux.
...
+You can customise some parameters, but the default values should suit most users.
```
hoechenberger, line 13: American spelling please :) I know some folks living outside the USA dislike this, but I'm aiming for consistency, and in computer-related stuff, American seems to be dominant 😅

README.md
```diff
@@ -4,3 +4,26 @@ Installers for MNE-Python for macOS, Windows, and Linux.
...
+Once completed, a conda evironment should be installed in your home directory.
```
hoechenberger, line 14: In fact, on Windows and Linux, users should be able to choose to install into whichever directory they want.

On the Mac, with the graphical installer, the directory is pre-defined (side note: it is also possible to create a `.sh` installer for the Mac that offers the same flexibility as the Linux installer, but I decided against this for now. Willing to reconsider if you think that was not a good decision)

At any rate, currently the graphical Mac installer dumps everything to `~/mne-python`. I want to change this to `~/mne-python/0.23.0` (or which ever version the user installs) to make it possible to have multiple versions installed in parallel (see #7). Do you think this is actually a good idea?

README.md
```diff
@@ -4,3 +4,26 @@ Installers for MNE-Python for macOS, Windows, and Linux.
...
+
+## Usage
+
+### MacOS
+
+Move to your home directory and activate the environment:
+
+```bash
+cd ~
+source mne-python/bin/activate
```
hoechenberger, line 25: Would it not be easier to do it without the `cd`?

`source ~/mne-python/bin/activate`
?

btw:
If users already have `conda` installed, they can also run
```bash
conda activate ~/mne-python
```
This will also produce a more obvious prompt prefix
(but I have to admit that asking users to first install Miniconda kind of goes against the entire idea of the installer)

hoechenberger (commented):
Thanks!! Left a few comments

README.md
rob-luke: > but I decided against this

Agree, simpler is best. If someone wants anything more they should use the traditional python install path and get all the flexibility they want.

> make it possible to have multiple versions installed in parallel (see #7). Do you think this is actually a good idea

Yes, this would be widely used with my colleagues. People like to chose a version at the start of a project and stick with it (unless bugs of course) until the project ends. And often there is multiple projects going at once. Long story short, multiple versions is good.

README.md
rob-luke: > asking users to first install Miniconda kind of goes against the entire idea of the installer

Agreed.

hoechenberger:
Hey @rob-luke, how should we proceed here? :)

rob-luke:
It seems ive broken the repo CI! Any other comments? Otherwise merge? I can test on linux next week add usage instructions there too in a follow up PR if you want.

rob-luke:
Looks like this is back on the cards? Let me know how I can help. Easy install would be HUGE win for me, I spend lots of my time helping people install.

hoechenberger:
Let's get this done next week!

rob-luke:
Hi @hoechenberger 

I see this initiative has just kicked off again. Great.
I just downloaded the latest release and followed the exact instructions from this PR and it worked as described.
I tested this on a Mac with PyCharm and JupyterLab.

Any additional testing you want me to do?

hoechenberger:
Thanks @rob-luke!

I'm now trying to tune some things and will reach out to you soon and ask for further help testing 😅😅😅

PR #23, opened 2022-01-11
Try to work around bugs in constructor

hoechenberger:


hoechenberger:
Seems to work for Linux so far, merging and will iterate from there.

PR #38, opened 2022-02-11
Add more packages

hoechenberger:
Several of those are still waiting to be merged into conda-forge, so this build will fail for now.

@larsoner @agramfort What's your opinion on including a set of packages commonly used by researchers in our field & scientists we know personally? I generally like the idea of including everything a user would ever need, as long as it doesn't cause us any pain. The installer packages are several 100 MBs in size already anyway, adding another 50 or 100 or even 200 doesn't make much of a difference, right? But where do we draw the line? And: do you have any suggestions for other packages that should be included? In theory, since we're using `conda`, we could even install actual tools like `git`. How far do we want to go? :)

cc @cbrnr @marsipu

### DIFF
--- a/recipes/mne-python_0.24/construct.yaml
+++ b/recipes/mne-python_0.24/construct.yaml
@@ -33,6 +33,8 @@ channels:
 specs:
   - mne ==0.24.1
   - mne-bids =0.9
+  - mne-connectivity =0.2
+  - mne-faster =1.0
   - mne-nirs =0.1
   - python =3.9
   - pip
@@ -50,11 +52,19 @@ specs:
   - ipyvtklink
   - darkdetect
   - qdarkstyle
+  # Helpful for statistics
+  - pingouin
+  # MRI viewer
+  - fsleyes
   # NeuroSpin needs the following packages
   - questionary
   - dcm2niix
   - neurodsp
   - bycycle
+  - fooof
+  - pycircstat
+  - pactools
+  - tensorpac
 
 condarc:
   channels:
--- a/recipes/mne-python_0.24/construct.yaml
+++ b/recipes/mne-python_0.24/construct.yaml
@@ -52,6 +52,9 @@ specs:
   - ipyvtklink
   - darkdetect
   - qdarkstyle
+  # Excel I/O
+  - openpyxl
+  - xlrd
   # Helpful for statistics
   - pingouin
   # MRI viewer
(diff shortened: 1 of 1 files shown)


cbrnr:
MNELAB would probably be very useful for many users, but it needs to be `pip` installed (because of `PySide6`).

hoechenberger:
@cbrnr Or can Qt5 and 6 be installed in parallel? Probably not?

cbrnr:
And yes, `PySide6` can be installed in parallel with `PySide2` and `PyQt5`, no problem there.

hoechenberger:
> Oh well, that's not going to happen then.

I linked to a PR where somebody's actively working on this

hoechenberger:
> And yes, `PySide6` can be installed in parallel with `PySide2` and `PyQt5`, no problem there.

We could install it via `pip` then, that would indeed be possible. We'd just have to be careful not to mess up half of our dependencies :)

cbrnr:
> I linked to a PR where somebody's actively working on this

I saw that. It would be great if this gets merged quickly.

> We could install it via pip then, that would indeed be possible. We'd just have to be careful not to mess up half of our dependencies :)

That should be doable. Basically you only install packages via pip that are not available in conda, i.e. `PySide6` and `mnelab`. I can figure out the exact `pip install` command, and the other dependencies can be installed via `conda` (`numpy`, `scipy`, `pyxdf`, etc.).

hoechenberger:
> That should be doable. Basically you only install packages via pip that are not available in conda, i.e. `PySide6` and `mnelab`. I can figure out the exact `pip install` command, and the other dependencies can be installed via `conda` (`numpy`, `scipy`, `pyxdf`, etc.).

Thanks! As it turns out in #39, it's not super trivial to do, but I still have some ideas how to make this work

agramfort:
no objection

larsoner:
+1 to include `openpyxl` for ease of reading excel spreadsheets (it's even needed for `pandas`, not sure if it gets installed by default with `pandas` or not)

hoechenberger:
Wasn't there also another package for writing XLSX?

larsoner:
No idea, I only ever write CSV...

hoechenberger:
Note to self: probably no need to use xlsxwriter as alternative to openpyxl; however xlrd should be included for reading XLS files

PR #39, opened 2022-02-11
Try installing MNELAB from PyPI

hoechenberger:
cc @cbrnr

Let's see how this goes, I have not tested this locally at all 😅

### DIFF
--- a/recipes/mne-python_0.24/construct.yaml
+++ b/recipes/mne-python_0.24/construct.yaml
@@ -55,6 +55,8 @@ specs:
   - dcm2niix
   - neurodsp
   - bycycle
+  - pip:
+    - mnelab =0.7
 
 condarc:
   channels:


hoechenberger:
Oh, bummer. `pip` installations don't appear to work with the `spec` section, only with an environment file? Need to investigate.

cbrnr:
I'd install with `--no-deps` like this:

`pip install mnelab PySide6 shiboken6 --no-deps`

The other packages should be available in `conda-forge`:

`matplotlib`, `numpy`, `mne`, `scipy`, `pyxdf`, `pyobjc-framework-Cocoa`

larsoner:
@hoechenberger can you please fork this repo, change your local `origin` to use your fork, and then reopen this PR from your fork? We should delete the `mnelab` branch from this repo

hoechenberger:
I think we can delete this without the need to re-open: currently there's not a trivial way to ship `pip` packages with the installer.

I mean, actually, there is:
We can include arbitrary files – e.g., wheels – and then install them via `pip` in the post-install script. But I'm not sure if this level of fiddling is worth it right now. I still hope we'll see Qt6 hit `conda-forge` within the next few months.

Speaking of which – there won't be an easy way to install Qt5 and Qt6 in parallel via `conda`. Will that break our neck?

PR #51, opened 2022-02-12
MAINT: Use concurrency

larsoner:
This should cancel in-progress builds when a PR pushes new commits

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,9 +24,6 @@ jobs:
 
     continue-on-error: true
 
-    # concurrency:
-    #   group: ${{ github.head_ref }}
-    #   cancel-in-progress: true
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -21,7 +24,7 @@ jobs:
 
     continue-on-error: true
 
-    # concurrency: 
+    # concurrency:
     #   group: ${{ github.head_ref }}
     #   cancel-in-progress: true
```
hoechenberger, line 26: I think we can remove these lines then!

hoechenberger (commented):
Super cool! Just one recommendation for a cleanup. Thanks @larsoner!

.github/workflows/build.yml
hoechenberger, line 29:
```suggestion
```

PR #52, opened 2022-02-13
Package MNE-Python 1.0 snapshot

### LINKED ISSUE #50: Update to 1.0?
Any reason all builds are currently named 0.24? It seems like they should be 1.0.dev0 or so, since that's what they need to work with...?

hoechenberger:
Fixes #50

cc @larsoner

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Build installer
         shell: bash -l {0}
         run: |
-          constructor recipes/mne-python_0.24
+          constructor recipes/mne-python_1.0
 
       - name: Run installer (macOS)
         if: ${{ runner.os == 'macOS' }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,25 +43,25 @@ jobs:
         if: ${{ runner.os == 'macOS' }}
         run: |
           echo `pwd`
-          installer -pkg MNE-Python-0.24.1_2-macOS_Intel.pkg -target CurrentUserHomeDirectory
+          installer -pkg MNE-Python-1.0.0_0-macOS_Intel.pkg -target CurrentUserHomeDirectory
 
       - name: Run installer (Linux)
         if: ${{ runner.os == 'Linux' }}
         run: |
           echo `pwd`
-          sh ./MNE-Python-0.24.1_2-Linux.sh -b
+          sh ./MNE-Python-1.0.0_0-Linux.sh -b
 
       # https://docs.anaconda.com/anaconda/install/silent-mode.html
       - name: Run installer (Windows)
         if: ${{ runner.os == 'Windows' }}
         shell: cmd  # Couldn't get PowerShell to properly perform install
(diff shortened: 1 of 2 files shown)


hoechenberger:
@larsoner I'm not updating Python to 3.10 here, as apparently with the huge stack of software we're now including, dependency resolution fails if we use Python 3.10. Hopefully this will be resolved by the time we do the first 1.0-followup release.

larsoner:
> I'm not updating Python to 3.10 here, as apparently with the huge stack of software we're now including, dependency resolution fails if we use Python 3.10. Hopefully this will be resolved by the time we do the first 1.0-followup release.

VTK has no 3.10 release, so we're probably stuck for a while. I think it's fine to be on 3.9 for now

hoechenberger:
@larsoner There are 3.10 builds of VTK on conda-forge. It's none of our MNE dependencies holding us back here, it's one of the other packages we're including in the installer, I just haven't narrowed down which one. But like you said, sticking with 3.9 should be okay for now

rob-luke:
> Would you know if the latest stable release of MNE-NIRS is compatible with MNE-Python 1.0

Not sure. I can check Thursday and release a new version if required (currently traveling).

rob-luke:
> Would you know if the latest stable release of MNE-NIRS is compatible with MNE-Python 1.0

Ok, I am back at my desk and can look at MNE-NIRS now. The main branch of MNE-NIRS is compatible with MNE-Python main, it is not compatible with 0.24. The current released version of MNE-NIRS is compatible with MNE-Python 0.24.

I usually wait until MNE-Python cuts a release, then I update the dependency of MNE-Python in MNE-NIRS to match the latest release. In other words, I would usually wait for MNE-Python to release 1.0. Then I will change the requirements in MNE-NIRS to require `mne_python >= 1.0` and cut a new release.

Does this answer your question @hoechenberger

hoechenberger:
Thanks @rob-luke, this helps!

rob-luke:
Let me know if you want me to do anything different @hoechenberger

hoechenberger:
Im really undecided how to go about this! Any chance you'd be able to join us in tomorrow's dev meeting?

rob-luke:
The next meeting is in the 26th at 3am AEST?

rob-luke:
Sorry, I won't make 3am (usually I can push myself, but 3am is rough :wink:).
If the release of MNE-Python is made on a week day, then I can ensure a new MNE-NIRS is released the following day (if that helps).

hoechenberger:
I don't think there's reason to stress out, we'll just have to find a solution that works best for everyone :)

hoechenberger:
> @larsoner I'm not updating Python to 3.10 here, as apparently with the huge stack of software we're now including, dependency resolution fails if we use Python 3.10. Hopefully this will be resolved by the time we do the first 1.0-followup release.

It seems that `dipy` is the culprit: it currently doesn't support Python 3.10.
https://github.com/conda-forge/dipy-feedstock/pull/37#issuecomment-1002335909

PR #57, opened 2022-02-16
Create signed macOS installer

hoechenberger:


hoechenberger:
Screenshot 2022-02-17 at 13 52 07

GuillaumeFavelier:
Awesome! I think this will make MacOS users really happy :)

hoechenberger:
I was going to start singing the windows installer next, but MS doesn't offer such Certs, and purchasing them from trusted 3rd parties costs about 400 USD annually. Bummer

larsoner:
Agreed that's not worth it

hoechenberger:
It's just super annoying because asking users to install software that the system clearly marks as "untrusted" seems like a very bad thing to do 😟 I asked a friend of mine who's a regular windows user and he says these types of dialogs are extremely rare these days:

image

larsoner:
Can you check to see if the MNE-CPP or MNE-SCAN folks have dealt with this problem? They might have a license we could use

marsipu:
> @marsipu Got any opinion on this?
Frankly from my personal view this wouldn't be unexpected, since most of the time when I installed scientific open source programs, this pops up and I got used to it. Since we are OpenSource, concerned users could just look into our code if they fear anything malicious.

hoechenberger:
> Can you check to see if the MNE-CPP or MNE-SCAN folks have dealt with this problem? They might have a license we could use

They simply distribute unsigned archive full of unsigned executables, not even checksums or anything … same for all platforms.

When I try to run any of their `.app` bundles on my Mac, this is what I get:

Screenshot 2022-02-17 at 20 42 05

Not good.

hoechenberger:
For our upcoming first public release of the installer, I'll at least try to distribute SHA256 hashes along with the installer files. I'll sign the hashes so their validity can be verified. Need to decide which types of certificates to use: X.509 or OpenPGP. I suppose verification via X.509 should work on all platforms without the need to install additional software.

PR #70, opened 2022-02-28
On macOS, install to /Applications/MNE-Python and set folder icon

hoechenberger:


hoechenberger:
Screen Shot 2022-02-28 at 10 52 24

PR #77, opened 2022-03-14
Use Python 3.10

hoechenberger:


larsoner:
I'm not sure what the problem is, but it seems wrong that there are no py310 wheels in the dipy 1.5 list here:

https://anaconda.org/conda-forge/dipy/files

larsoner:
https://github.com/conda-forge/dipy-feedstock/pull/39#issuecomment-1066920757

larsoner:
@hoechenberger can you try removing `dipy` from the installer in this PR temporarily? That way we can see if https://github.com/conda-forge/dipy-feedstock/pull/40 is likely to fix things here or if there are other things that we need to figure out...

hoechenberger:
@larsoner It's actually a dependency of `mne` but I think we can force-exclude it, yes. Will try…

hoechenberger:
Ok there's still a problem, I'm canceling the CI run

hoechenberger:
@larsoner Let's wait until the `dipy` package is up and then retry, I'm not too sure about the `exclude` parameter I've just added.

If things still fail with a fresh `dipy`, we can start to dig deeper.

hoechenberger:
@larsoner Ah, in fact, I believe `exclude` only means that those packages won't be shipped with the installer, however it does not mean that they're excluded from the dependency resolution procedure that has to run before the installers are actually being created. So this was not the right approach anyway.

larsoner:
The mne-base approach should work though, right? I can make a PR if you are working on instructions

hoechenberger:
@larsoner Yes you can do it, but I don't think it's time-efficient … I'd just wait until the new `dipy` packages are up and then restart CI here

But feel free to do it if you're impatient :)

recipes/mne-python_1.0/construct.yaml
```diff
@@ -56,9 +56,9 @@ specs:
   # TODO: ⛔️ ⛔️ ⛔️ DEV BUILDS START: CHANGE BEFORE RELEASE! ⛔️ ⛔️ ⛔️
   - mne =1.0*=*20220308
   - mne-installer-menus =1.0*=*20220308
-  - mne-bids =0.10*=*20220213
   # TODO: ⛔️ ⛔️ ⛔️ DEV BUILDS STOP: CHANGE BEFORE RELEASE! ⛔️ ⛔️ ⛔️
 
+  - mne-bids =0.10
   - mne-connectivity =0.3
   - mne-faster =1.1
   - mne-nirs =0.2
```
larsoner, line 64: This will need to be 0.2.1 as well

hoechenberger:
@larsoner WE'RE GREEN 💚

PR #78, opened 2022-03-14
MAINT: Fix conditionals for PRs

larsoner:
Let's see if dipy 3.10 is likely to fix our problems or if there is something else we can know in advance we need to investigate

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,13 +100,19 @@ jobs:
         run: |
           patch $CONDA_PREFIX/lib/python3.9/site-packages/constructor/osxpkg.py ./assets/constructor_macOS.patch
 
+      - name: Patch config (macOS pull request)
+        if: ${{ runner.os == 'macOS' && github.event_name == 'pull_request' }}
+        shell: bash -l {0}
+        run: |
+          sed -i "s/9779L28NP8//" recipes/mne-python_1.0/construct.yaml
+
       - name: Build installer
         shell: bash -l {0}
         run: |
           PYTHONUTF8=1 constructor recipes/mne-python_1.0  # enforce UTF-8 encoding when reading files even on Windows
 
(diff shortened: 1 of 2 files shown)


larsoner:
... I'll push a commit to take care of it

hoechenberger:
@larsoner You'd have to patch `construct.yml` as well if you don't fetch the certs … as they are mentioned in this config file
I don't think it's worth adding this level of complexity!

hoechenberger:
@larsoner Let's keep the signing but disable the notarization in PRs, WDYT?

larsoner:
I am not entirely clear on the differences here, but I would not allow PRs to use any steps that require secrets. If you can sign without secrets and need secrets to notarize, then I agree that seems reasonable

hoechenberger:
Like I said, the only way to skip signing is by patching `construct.yaml` as this file specifies IDs of signing keys. If the keys cannot be found, the build fails.

.github/workflows/build.yml
```diff
@@ -100,13 +100,19 @@ jobs:
         run: |
           patch $CONDA_PREFIX/lib/python3.9/site-packages/constructor/osxpkg.py ./assets/constructor_macOS.patch
 
+      - name: Patch config (macOS pull request)
+        if: ${{ runner.os == 'macOS' && github.event_name == 'pull_request' }}
+        shell: bash -l {0}
+        run: |
+          sed -i "s/9779L28NP8//" recipes/mne-python_1.0/construct.yaml
```
hoechenberger, line 107: macOS ;))

```suggestion
          sed -i "" "s/9779L28NP8//" recipes/mne-python_1.0/construct.yaml
```

hoechenberger:
@larsoner This is nicer though: https://github.com/hoechenberger/mne-installers/pull/78#discussion_r826290194

hoechenberger (approved):

PR #79, opened 2022-03-14
WIP: Try base

larsoner:
We'll have to ignore macOS until #78 is merged. But in the meantime it looked like we did have some conflicts/problems before I started working on that stuff. So let's get the ball rolling again...


larsoner:
> @larsoner Maybe if you reset the timeout to 15 min we might see a little bit more ￼

Looking at your other PR, I don't think it will help. It just times out later, unclear which package is which. I want it to work, or fail ASAP. I'm going to start bisecting the reqs to see if I can find the problem...

larsoner:
(In other words, this should fail in about 10 minutes each time I push, and I should only need ~8 steps to bisect the problematic package. If I'm vigilant this is ~80 minutes. By contrast your PR that tries to add 3.10 directly takes ~60 minutes to fail and the log is still not very useful, as far as I can tell...)

larsoner:
... of course I didn't count and removed more than half the packages, so if I'm unlucky it will take me more than 8 steps :)

hoechenberger:
FYI mamba provides MUCH clearer summaries of dep resolution fails. However mamba doesn't seem to play with labels for me (we need those to install the MNE dev package)

hoechenberger:
This doesn't make sense ... I've definitely installed MNE-base with Py3.10 on my system last week or so. Will have a look shortly (currently afk)

larsoner:
>  I've definitely installed MNE-base with Py3.10 on my system last week or so. Will have a look shortly (currently afk)

In `main` you don't currently just install even just `mne` but a bunch of other stuff like `fsleyes` etc. It's almost certainly one of those causing the problem. I'll remove them bit by bit and figure out the problem(s), then try restoring the `mne` (non-`base`) deps other than `dipy`. Hopefully that works...

larsoner:
Okay it was green, so it was the extra deps. Starting to put them back now...

hoechenberger:
My bet would be on `dask` or `distributed` :)

larsoner:
My bet is on some smaller package that is less well maintained -- those are presumably used by a lot of people so the pressure for 3.10 would be high. But we'll see!

hoechenberger:
@larsoner It's questionary:

https://github.com/conda-forge/questionary-feedstock/blob/5415f348301002213ed104738f03bd262e3f9588/recipe/meta.yaml#L23

(and maybe something else too, but this one is definitely one of our issues)

hoechenberger:
@larsoner I've opened a PR upstream at the `questionary` feedstock, https://github.com/conda-forge/questionary-feedstock/pull/5

larsoner:
@hoechenberger this is also ready for review/merge

hoechenberger (approved):

hoechenberger:
Thanks @larsoner, amazing work!

PR #80, opened 2022-03-15
Try to import all packages after installation in CI

hoechenberger:


hoechenberger:
Needs `autoreject` 0.3.1.

https://github.com/autoreject/autoreject/issues/262#issuecomment-1067700459

larsoner:
So `autoreject` and updated installation instructions are all we need at this point, right?

If so, maybe we should merge this as-is and transfer the repo to mne-tools. WDYT? If so, can you make me an owner of the repo? Then I can transfer it.

hoechenberger:
Yes to everything you said. 

Will do tonight. Will also share my draft of the install docs.

hoechenberger:
@larsoner I've transferred ownership to `mne-tools`. LMK if this doesn't give you all the permissions you need!

PR #81, opened 2022-03-15
ENH: Add mne-kit-gui

larsoner:
I expect this *not* to work because `mayavi` will probably break everything, but we'll see!

@hoechenberger to be safer it might be good to have a few other minimal tests that we run for the installers to make sure our main GUI frameworks are not broken (I expect these are the most likely things to be broken):

- [x] `matplotlib`: just do `plt.subplots()` or something and make sure it works (#83)
- [x] `matplotlib`: set the backend to QtAgg, make sure it works (#83)
- [x] `pyvistaqt`: test the Qt backend by opening a BackgroundPlotter, adding a mesh, and taking a screenshot (#83)
- [x] `pyvistaqt`: test notebook backend the same was as Qt (#83)
- [x] bump back to Python 3.10
- [x] wait for https://github.com/conda-forge/mayavi-feedstock/pull/59 to land and be uploaded
- [x] merge #83
- [x] rebase this PR
- [x] `mne-kit-gui` (should be done in this PR): open an instance of the coreg gui
- [x] wait for https://github.com/conda-forge/pyface-feedstock/pull/30

This sort of stuff will be especially useful as we try to add support for MNELAB, which uses Qt6 (?), and add abstractions like https://github.com/mne-tools/mne-python/pull/10430.

If you agree these would be useful, I'll open a separate PR for the first four items above. It'll probably require a little bit of CI tweaking to get OpenGL working for the pyvistaqt tests, but hopefully not too much.

### DIFF
--- a/recipes/mne-python_1.0/construct.yaml
+++ b/recipes/mne-python_1.0/construct.yaml
@@ -70,7 +70,7 @@ specs:
   - mne-kit-gui ~=1.0.1
   - autoreject ~=0.3
   # Python
-  - python >=3.9
+  - python >=3.10
   - pip
   - conda
   - mamba
--- a/recipes/mne-python_1.0/construct.yaml
+++ b/recipes/mne-python_1.0/construct.yaml
@@ -70,7 +70,7 @@ specs:
   - mne-kit-gui ~=1.0.1
   - autoreject ~=0.3
   # Python
-  - python =3.10
+  - python >=3.9
   - pip
   - conda
   - mamba
--- a/recipes/mne-python_1.0/construct.yaml
+++ b/recipes/mne-python_1.0/construct.yaml
@@ -67,6 +67,7 @@ specs:
   - mne-rsa ~=0.6.0
   - mne-microstates ~=0.3.0
   - mne-ari ~=0.1.0
+  - mne-kit-gui ~=1.0.1
   - autoreject ~=0.3
   # Python
   - python =3.10
--- a/tests/test_gui.py
+++ b/tests/test_gui.py
@@ -42,3 +42,16 @@ raw = mne.io.RawArray(np.zeros((1, 1000)), mne.create_info(1, 1000., 'eeg'))
 fig = raw.plot()
 fig.close()
 assert 'PyQtGraphBrowser' in repr(fig), repr(fig)
+
+# mne-kit-gui
+from pyface.api import GUI  # noqa
+import mne_kit_gui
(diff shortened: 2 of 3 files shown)


hoechenberger:
@larsoner What was the blocker here for Python 3.10?

larsoner:
> What was the blocker here for Python 3.10?

Locally I get a failure with
```
$ conda create -c conda-forge -n mayavi "python>=3.10" mayavi
```
Looking at the conda-forge files page they don't have 3.10. Looks like someone would have to fix https://github.com/conda-forge/mayavi-feedstock/pull/57 :(

larsoner:
Looks like a bug that should be fixed in TraitsUI 7.3+. Will have to check on that feedstock...

larsoner:
Okay, now pyface... https://github.com/conda-forge/pyface-feedstock/pull/30

hoechenberger:
> Okay, now pyface... [conda-forge/pyface-feedstock#30](https://github.com/conda-forge/pyface-feedstock/pull/30)

Wow, so this wasn't being properly maintained anymore..?

What was the issue with TraitsUI?

larsoner:
TraitsUI needed to be updated to include a fix I made for it back in October. Updating TraitsUI requires updating pyface -- this is a guess really but it makes sense. I don't know how carefully they test inter-version compat at their (traits/traitsui/pyface) end, but they have some inter-dependencies that tend to cause these sorts of problems.

And yes traitsui and pyface were 2-3 versions behind, I'm adding myself as a maintainer and setting up automerge...

larsoner:
> GitHub Actions is experiencing degraded performance. We are still investigating and will provide an update when we have one.

[Argh](https://www.githubstatus.com/)

recipes/mne-python_1.0/construct.yaml
```diff
@@ -67,9 +67,10 @@ specs:
   - mne-rsa ~=0.6.0
   - mne-microstates ~=0.3.0
   - mne-ari ~=0.1.0
+  - mne-kit-gui ~=1.0.1
   - autoreject ~=0.3
   # Python
-  - python =3.10
+  - python >=3.10
```
hoechenberger, line 73:
```suggestion
  - python ~=3.10.0
```

larsoner:
The last one came back all green! I'll go ahead and merge this assuming it comes back green again. Then MNE-KIT-GUI can even be included in the release! I assumed Mayavi would make this impossible...

PR #83, opened 2022-03-15
TST: Test GUI frameworks

### LINKED ISSUE #84: Move test-related files from /assets to a new, dedicated directory, e.g., /tests
The original idea of `/assets` was once to only include stuff that will be used by the installers / will be bundled with them.

cc @larsoner

larsoner:
I expect this to fail because we need something like `nbclient`. It should be light enough to add if we do...

Closes #84


hoechenberger:
> I expect this to fail because we need something like `nbclient`. It should be light enough to add if we do...

Just add it, then :) with a comment that we need it for testing only

.github/workflows/build.yml
```diff
@@ -251,11 +254,13 @@ jobs:
           popd
 
           # Run MNE's sys_info
-          xvfb-run mne sys_info
+          mne sys_info
```
hoechenberger, line 257: Will this correctly report pyvista?

larsoner:
... it did not fail, at least on Windows and Linux. macOS failed because we can't switch from the MacOSX backend to Qt, so I separated out the default from non-default for that OS. Ready for review/merge from my end @hoechenberger !

hoechenberger (approved):

hoechenberger:
> macOS failed because we can't switch from the MacOSX backend to Qt

Why's that?

Feel free to merge once we're green!

larsoner:
> Why's that?

Probably some protection against using multiple frameworks simultaneously in a given session. I know it can cause problems sometimes... I've lost the log now, but `matplotlib.use('QtAgg', force=True)` failed with an error message saying that it couldn't switch because `MacOSX` was already in use.

tests/test_gui.py
```diff
@@ -0,0 +1,43 @@
...
+fname = this_path / 'test.png'
+mne.viz.set_3d_backend('pyvista')
+fig = mne.viz.create_3d_figure((400, 400))
+plotter = fig.plotter
+plotter.add_orientation_widget(pyvista.Cube())  # Old test without color='b'
+plotter.add_mesh(pyvista.Cube())
+if fname.is_file():
+    os.remove(fname)
+assert not fname.is_file()
+plotter.screenshot(fname)
```
larsoner, line 32:
```suggestion
plotter.render()
plotter.screenshot(fname)
```

Issue #94, opened 2022-03-19
On macOS, install to /Applications/MNE-Python and drop the version numbers from .app bundle names


hoechenberger:
Also switch from naming style `MNE Spyder` to `Spyder (MNE)`

see #90

hoechenberger:
Dropping version numbers needs to be addressed upstream in https://github.com/conda-forge/mne-feedstock

PR #97, opened 2022-03-19
On macOS, install to /Applications/MNE-Python

hoechenberger:
First part of #94


hoechenberger:
@larsoner Appears we cannot have spaces in the installation prefix :( https://github.com/conda/constructor/pull/449

larsoner:
> How about we simply omit all version numbers on macOS… We have the System Info shortcut that tells users which version of MNE-Python they're using. Same for the Prompt. WDYT?

Works for me. By the time 1.1 is out, maybe spaces will be supported and we can add it back in...

PR #105, opened 2022-03-21
Use MNE-Python 1.0.0 stable

hoechenberger:
cc @larsoner

Will fail until the new packages are on the CDN

### DIFF
--- a/recipes/mne-python_1.0/construct.yaml
+++ b/recipes/mne-python_1.0/construct.yaml
@@ -43,17 +43,16 @@ menu_packages:
 
 channels:
   - conda-forge
-  - conda-forge/label/mne_dev
+  # - conda-forge/label/mne_dev
   # - conda-forge/label/mne-bids_dev
 
 specs:
   # MNE ecosystem
 
   # TODO: ⛔️ ⛔️ ⛔️ DEV BUILDS START: CHANGE BEFORE RELEASE! ⛔️ ⛔️ ⛔️
-  - mne =1.0*=*20220319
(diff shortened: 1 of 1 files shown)


larsoner:
FYI I've put this in step 17 in https://github.com/mne-tools/mne-installers/pull/105

larsoner:
... I mean https://github.com/mne-tools/mne-python/wiki/How-to-make-a-release

larsoner:
@hoechenberger I say we merge this once green so that we have some working installers, then update the URLs on the homepage to use them. That takes away any time pressure from release.

Then we merge https://github.com/conda-forge/mne-feedstock/pull/88, rebuild in another PR here, and iterate until they work on all platforms. Once they do, we can update the installer pages again. WDYT?

hoechenberger:
Yes +100 for me

hoechenberger:
I'm off for the rest of the day, do what you think is best. We'll send out the announcement tomorow, right? @agramfort

larsoner:
I tested the draft pre10 on all three platforms and things appeared to work, so I cut a new `v1.0.0` non-draft, non-prerelease version (+tag). Once the assets show up, I'll update the installation instructions with `[circle deploy]` for both. Then I think things will be all set for release announcements (I edited the google doc @agramfort) tomorrow

hoechenberger:
Amazing. Thanks so much @larsoner 

I'm also feeling comfortable with tuning and improving the installer whenever an idea or an issue pops up. In the worst case, users can always go back to using our conda install instructions if the installer doesn't work in their system until we provide a fix.

PR #111, opened 2022-03-30
Use mne-qt-browser 0.2.6

### LINKED ISSUE #110: Release new installers with updated mne-qt-browser 0.2.6

hoechenberger:
Fixes #110


hoechenberger:
We have a problem related to `emd`. If I cannot resolve this quickly, I will temporarily drop this package from the installers to allow us to quickly do a release with the updated `mne-qt-browser`.

PR #112, opened 2022-03-31
Remove Jinja2 pinning again

hoechenberger:
The numpydoc package on conda-forge has been updated,
so now we shouldn't get into trouble because
of unsatisfiable Jinja2 dependencies again.


hoechenberger:
Actually this doesn't guarantee we pull in the latest numpydoc, so we either have to be pinning Jinja2 for now or numpydoc (to a specific build), which would then implicitly pin Jinja2.

So I'm closing this for now and will make a release.

PR #113, opened 2022-04-01
WIP: Bash entry for windows, normalize names

larsoner:
1. it seems like we should use `_post2` rather than `_2` in our names as it's more standard (and it's what we're calling our releases anyway)
2. This uses the `_2` conda builds, so it should include a Bash entry on Windows now.

@hoechenberger if you're happy with this, could you merge and cut a release draft, then you or I could test on Windows? I'd encourage you to try VirtualBox again, the ~20 GB [Windows test VM](https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/) worked pretty well for me there in testing!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -217,7 +217,7 @@ jobs:
 
           echo "Checking we have all .app bundles in /Applications/MNE-Python:"
           ls -d /Applications/MNE-Python/*.app
-          test `ls -d /Applications/MNE-Python/*.app | wc -l` -eq 5
+          test `ls -d /Applications/MNE-Python/*.app | wc -l` -eq 3
 
           echo "Checking that the custom icon was set on the MNE folder in /Applications/MNE-Python"
           test -f /Applications/MNE-Python/Icon$'\r'
@@ -250,7 +250,7 @@ jobs:
           ls -l
           echo "Checking for existence of .desktop files:"
           ls MNE-Python*.desktop
-          test `ls MNE-Python*.desktop | wc -l` -eq 5
+          test `ls MNE-Python*.desktop | wc -l` -eq 3
           echo ""
 
           # … and patched to work around a bug in menuinst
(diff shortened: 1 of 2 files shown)


hoechenberger:
No, not happy. The only reason I append _number is because _YYYYMMDD got too long when I tried it early ob. But I think we have since changed the string before that and we can try and see if adding the date again would work. 

I don't want to bother users with highly technical stuff like postwhatever, while still indicating which revision the installers have.

larsoner:
> I want some sort of counter, or the date (which I'd prefer)

I think a counter might be better, for the same reasons as above: 1) users probably don't care about this technicality, and 2) it makes the name even longer. I'll revert to `_2`

larsoner:
Hooray, a conflict on Windows :(

The other unsolved issue here is why I had to `-eq 3` here for `.desktop` and `.app` rather than `-eq 5` on `main`. https://github.com/mne-tools/mne-installers/pull/114 should tell us if it's really from this PR or from (not) checking the exit codes of all commands.

larsoner:
https://github.com/mne-tools/mne-installers/pull/114 should be merged first, then I'll rebase this

hoechenberger:
> I guess that's a problem where `mne-feedstock` does not build the `mne-installer-menus` target.

Bumping the build number **should** enforce a rebuild though … And in fact, the package was rebuilt: https://anaconda.org/conda-forge/mne-installer-menus/files

So I'm not sure where the conflict is coming from …

Oh, yeah, I might have an idea. The package is `noarch` so it's only built on Linux. We could make it non-`noarch` to avoid such breakage in the future, I think. But it really shouldn't happen in the first place :(

larsoner:
Let's try running now, maybe it was some temporary resolution error.

If it doesn't work, we should just revert https://github.com/conda-forge/mne-feedstock/pull/89 so that we can make the 1.0.1 installers.

PR #114, opened 2022-04-01
MAINT: Raise errors

larsoner:
It's weird that I had to change the `-eq 5` to `-eq 3` in https://github.com/mne-tools/mne-installers/pull/113. I want to check to see if it's from the `bash -e` arg, which errors out immediately on a problem (rather than allowing other commands to continue and override the exit code with something that potentially succeeds).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -318,7 +318,7 @@ jobs:
       - name: Export frozen environment definition (Windows)
         if: ${{ runner.os == 'Windows' }}
         run: |
-          $MneRoot = "$env:UserProfile\mne-python\${MNE_INSTALLER_VERSION}"
+          $MneRoot = "$env:UserProfile\mne-python\$env:MNE_INSTALLER_VERSION"
           . "$MneRoot\shell\condabin\conda-hook.ps1"
           conda activate $MneRoot
           mamba list --json > MNE-Python-$env:MNE_INSTALLER_VERSION-Windows.env.json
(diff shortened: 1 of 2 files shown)


larsoner:
@hoechenberger since this has come back green, I propose 

1. We merge this PR, as it "just" ensures that we use `-e` in BASH and use a programatically determined version. Then
2. We can isolate any build issues with the `_2`/`post2` conda-forge build in #113 after a rebase/merge with `main`

hoechenberger:
@larsoner Feel free to merge if you think this is ready.

Minor request:
Could you please not prefix the PR titles with `MAINT` etc? Since We're auto-compiling the changelog, these prefixes then quickly cause an eyesore. If you insist on adding this kind of metadata, could we agree to use actual tags instead?

larsoner:
> Could you please not prefix the PR titles with MAINT etc? Since We're auto-compiling the changelog, these prefixes then quickly cause an eyesore. If you insist on adding this kind of metadata, could we agree to use actual tags instead?f you insist on adding this kind of metadata, could we agree to use actual tags instead?

I usually just try to follow what most scientific python packages I work on (numpy, scipy, scikit-learn, mne-python, etc.) do / have done for years. The more we make each repo an exception with its own rules, the more rules people need to learn. I'd rather have to tweak the release notes at release time a bit (or live with a minor eyesore) than cause all contributors to have to use and remember a non-standard scheme.

hoechenberger:
> I'd rather have to tweak the release notes at release time a bit (or live with a minor eyesore) than cause all contributors to have to use and remember a non-standard scheme.

Can you not simply leave the prefixes out, then? Nobody really cares about them anyway?

PR #115, opened 2022-04-02
TST: Minimal error

larsoner:
https://github.com/mne-tools/mne-installers/pull/114 shows an error, let's see if we can *really* narrow it down to `bash -e` making a (good) difference to show an existing (bad) error. If it doesn't, it's possible that this is somehow related to the version extraction (?).


larsoner:
Looks like I messed two `_2`'s in #114, I'll optimistically close this under the assumption that this will fix that PR

PR #118, opened 2022-04-15
MAINT: Bump to 1.0.2

larsoner:
Will fail until https://github.com/conda-forge/mne-feedstock/pull/93 lands and then hits CDNs later today

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -196,6 +196,29 @@ jobs:
         run: |
           .\MNE-Python-%MNE_INSTALLER_VERSION%-Windows.exe /S /InstallationType=JustMe /AddToPath=1
 
+      - name: Export frozen environment definition (macOS)
+        if: ${{ runner.os == 'macOS' }}
+        run: |
+          source /Applications/MNE-Python/.mne-python/bin/activate
+          mamba list --json > MNE-Python-${MNE_INSTALLER_VERSION}-macOS_Intel.env.json
+          cat MNE-Python-${MNE_INSTALLER_VERSION}-macOS_Intel.env.json
+
+      - name: Export frozen environment definition (Linux)
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          source ~/mne-python/${MNE_INSTALLER_VERSION}/bin/activate
+          mamba list --json > MNE-Python-${MNE_INSTALLER_VERSION}-Linux.env.json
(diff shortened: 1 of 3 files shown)


recipes/mne-python_1.0/construct.yaml
```diff
@@ -56,8 +56,8 @@ specs:
   - jinja2 <3.1,>=2.10  # Remove once https://github.com/conda-forge/numpydoc-feedstock/pull/17 has been merged
   # TODO: ⛔️ ⛔️ ⛔️ REMOVE ASAP STOP ⛔️ ⛔️ ⛔️
 
-  - mne =1.0.1=*_0
```
larsoner, line 59: Looks like our `1.0.1_1` installer was actually using build 0 :(

@hoechenberger do you see a way to add a test that would have caught this? Can we `grep` the output json for example to make sure that the `mne` and `mne-installer-menus` that are used have the correct `version`?

larsoner:
Okay good now it failed, which it should  (no way 1.0.2_0 has hit CDNs yet). I found the bug only because it succeeded the first time incorrectly!

recipes/mne-python_1.0/construct.yaml
larsoner: Oh never mind I think we're okay on the 1.0.1 actually. But it still would have been nice to catch this error in this PR!

larsoner:
@hoechenberger FYI I think I have a test almost working, will push...

.github/workflows/build.yml
```diff
@@ -196,6 +196,29 @@ jobs:
         run: |
           .\MNE-Python-%MNE_INSTALLER_VERSION%-Windows.exe /S /InstallationType=JustMe /AddToPath=1
 
+      - name: Export frozen environment definition (macOS)
```
larsoner, line 199: This whole block is verbatim copy-paste just to make sure the `.json` is available before the testing blocks

larsoner:
[Got it!](https://github.com/mne-tools/mne-installers/runs/6041352473?check_suite_focus=true):
```
AssertionError: mne: got '1.0.1_0' != want '1.0.2_0'
```
Now let's see if it's fixed when I set the version properly in the version spec (and if the content is on the CDNs)

larsoner:
@hoechenberger ready for review/merge from my end

hoechenberger (approved):
Wow this is amazing!!

PR #120, opened 2022-04-20
Include pqdm

### LINKED ISSUE #119: Add pqdm
cc @agramfort

hoechenberger:
Fixes #119

### DIFF
--- a/recipes/mne-python_1.0/construct.yaml
+++ b/recipes/mne-python_1.0/construct.yaml
@@ -1,4 +1,4 @@
-version: 1.0.2_0
+version: 1.0.2_1
 name: MNE-Python
 company: MNE-Python Developers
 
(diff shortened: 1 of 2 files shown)


larsoner:
Okay the test complains now:
```
AssertionError: mne: got '1.0.2_0' != want '1.0.2_1'
```
This is because it's comparing the `mne-feedstock version_build` to the `version:` from the YAML, which no longer matches.

One option is to add a `feedstock_version: 1.0.2_0` or so to the YAML and do the test comparison with that, but that doesn't seem as helpful because it's another thing to change (and if you forget, the test is not useful).

A better change is probably to use a version scheme that includes the feedstock version, feedstock build, and mne-installer post, like this would be `version: 1.0.2_0_1` because it's the `feedstock-version_feedstock-build_installer-post`. The last release would have been `1.0.2_0_0` for example, and this would tick to `1.0.2_0_1`.

We could even start titling releases this way, maybe with a `v` prefix like `v1.0.2_0_1` if we cut a release after this PR. Then any automated process we use someday (maybe) to update MNE-Python docs could use the version number directly from the tag and plug it in the URL (minus the `v` prefix).

hoechenberger:
I will take a look / think about it 👍

larsoner:
Might as well take advantage of build 1 to include edflib-python now, too https://github.com/conda-forge/mne-feedstock/pull/94

larsoner:
... and soon build number 2 from https://github.com/conda-forge/mne-feedstock/pull/95

larsoner:
@hoechenberger don't forget about this one!

larsoner:
Should we add `edflib-python` even though it should now be pulled in by the `mne` dependency? Not sure if we want them to be listed both places or just allow dependencies to be pulled in automatically, but it looks like we tend to list both places...

hoechenberger:
> but it looks like we tend to list both places

Really? I think we shouldn't … unless we try to pin to a specific version here.
We already have so many duplications and complications going on, I'd like to avoid another set of those if we can :)

larsoner:
I thought there was overlap but looking at it I don't see much

https://github.com/conda-forge/mne-feedstock/blob/main/recipe/meta.yaml

So I guess what's left is resolving how we want to handle the versions. I think [my preference](https://github.com/mne-tools/mne-installers/pull/120#issuecomment-1103897718) is for

> use a version scheme that includes the feedstock version, feedstock build, and mne-installer post

PR #125, opened 2022-05-23
WIP: Add support for local builds

larsoner:
- [x] Fix bug where we were patching 3.9 even though we've mode to 3.10 (unclear to me why/how this hasn't created an error yet)
- [x] Move most calls to scripts
- [x] Add `assets/current_version.txt` so that `1.0` appears in only two places now (that file and `mne-python_1.0/recipes/construct.yaml` (i.e., DRY)
- [x] Start skeleton `build_local.sh` to replicate remote build
- [x] Add to README what to run locally to test/build
- [ ] Get CIs green
- [ ] Get it building+testing properly locally on Linux
- [ ] Get it building+testing properly on macOS (ideally M1, but at least Intel; I have access to both)


README.md
```diff
@@ -5,3 +5,12 @@ Installers for MNE-Python for macOS, Windows, and Linux.
...
 Please visit [the installers section of the MNE documentation](https://mne.tools/dev/install/installers.html) for instructions on how to use them.
+
+## Development
+
+Locally, installers can be built using `tools/build_local.sh`. This has only been tested on Linux and macOS M1.
+
+Requirements:
+- `conda` with `constructor` (`conda install -c conda-forge constructor`)
+
+      WARNING: Your constructor will be patched on macOS.
```
hoechenberger, line 16: This is not entirely correct, the requirements are listed here:

https://github.com/mne-tools/mne-installers/blob/c98e645690356ef3c23f1ff61c982b2db04e4744/environment.yml#L3-L11

We rely on a fork of `constructor` and `menuinst`

larsoner: ... and then on top of that there is a `patch` call, right?

hoechenberger: yes 😅😅😅😅😅

larsoner:
Okay this one is good to go @hoechenberger

recipes/mne-python_1.0/construct.yaml
```diff
@@ -33,8 +35,8 @@ post_install: ../../assets/post_install_windows.bat     # [win]
 
 # Create signed macOS .pkg installer
 installer_type: pkg                     # [osx]
-signing_identity_name: 9779L28NP8       # [osx]  Used for productsign
-notarization_identity_name: 9779L28NP8  # [osx]  Used for codesign
+signing_identity_name:                  # [osx]  Used for productsign
+notarization_identity_name:             # [osx]  Used for codesign
```
hoechenberger, line 39: This should be reverted for signing on CI, no?

tools/run_constructor.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -ef
+
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]:-$0}"; )" &> /dev/null && pwd 2> /dev/null; )";
+source ${SCRIPT_DIR}/extract_version.sh
+export PYTHONUTF8=1
+# enforce UTF-8 encoding when reading files even on Windows
```
hoechenberger, line 6: I think these two lines ought to be swapped :)

recipes/mne-python_1.0/construct.yaml
larsoner: See changes in CI. Rather than always have it in and remove for pull requests, I do the opposite and always have it blank but add it for non-PRs (main/release builds). This makes it work better for local builds as a useful side effect.

hoechenberger:
Thanks @larsoner!

PR #126, opened 2022-05-25
ENH: Build for M1

### LINKED ISSUE #48: Build on Apple Silicon
~~This is currently blocked by https://github.com/actions/runner/issues/805~~

~~*EDIT: Now blocked by https://github.com/actions/virtual-environments/issues/2187*~~

Waiting on:

- [x] https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2908 to be merged
- [x] dcm2niix: waiting on https://github.com/conda-forge/dcm2niix-feedstock/pull/17
- [x] mne-kit-gui: waiting on mayavi to appear on CDN
- [ ] spyder: waiting on CDN
- [x] fsleyes: waiting on https://github.com/conda-forge/fsleyes-feedstock/pull/90
- [x] sleepecg: waiting on CDNs
- [x] mne-installer-menus: Fixed by bumping build number to release that doesn't depend on Spyder
- [x] https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2910
- [x] openneuro-py
- [x] get build working locally on an M1 machine
- [x] get build cross-compile working locally on a non-M1 machine using [something like](https://github.com/mne-tools/mne-installers/issues/48#issuecomment-1136409419):
    ```
    constructor --platform=osx-arm64 --conda-exe=path/to/conda-standalone
    ```
- [x] add ARM to build matrix using command that worked locally
- [ ] merge PR
- [ ] cut a release
- [ ] test that the draft release installer actually works on M1

Status: https://conda-forge.org/status/#current_migrations

larsoner:
- [x] Add suggested path for building macos-arm64 on Intel
- [x] Add CI job
- [x] Fix `_Intel.pkg` naming
- [x] Ensure it's packaged properly (add some basic tests)
- [x] Get CI job green
- [x] Remove `[not arm64]` selectors once they're available

cc @hoechenberger currently I get conflicts when I try this locally on macOS Intel with the following on this branch (which passes the _ARG to run_installers.sh:
```console
$ source ./tools/setup_m1_crosscompile.sh
$ ./tools/run_installers.sh
```
Feel free to try it. I'm going to try doing a native run on my M1 machine with these `[not arm64]` selectors as well.

Closes #48

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
     steps:
       # Based on https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development
       - name: Install code-signing certificates (macOS)
-        if: ${{ runner.os == 'macOS' && github.event_name != 'pull_request' && matrix.arch == 'native'}}
+        if: ${{ runner.os == 'macOS' && github.event_name != 'pull_request' }}
         env:
           APPLICATION_CERT_BASE64: ${{ secrets.APPLE_APPLICATION_CERT_BASE64 }}
           APPLICATION_CERT_PASSWORD: ${{ secrets.APPLE_APPLICATION_CERT_PASSWORD }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,12 +121,12 @@ jobs:
         if: ${{ matrix.arch == 'arm64' }}
         shell: bash -el {0}
         run: |
-          conda install -c napari/label/bundle_tools conda-standalone
-          echo "PLATFORM_ARG=\"--platform=osx-arm64\"" >> $GITHUB_ENV
-          echo "EXE_ARG=\"--conda-exe=${CONDA_PREFIX}/standalone_conda/conda.exe\"" >> $GITHUB_ENV
+          source ./tools/setup_m1_crosscompile.sh
+          echo "EXE_ARG=${EXE_ARG}" >> $GITHUB_ENV
+          echo "PLATFORM_ARG=${PLATFORM_ARG}" >> $GITHUB_ENV
 
       - name: Patch config (macOS pull request)
-        if: ${{ runner.os == 'macOS' && github.event_name != 'pull_request' && matrix.arch == 'native' }}
+        if: ${{ runner.os == 'macOS' && github.event_name != 'pull_request' }}
         shell: bash -el {0}
         run: |
           sed -i "" "s/_name: *# \[osx\]/_name: 9779L28NP8  # \[osx\]/" ${RECIPE_DIR}/construct.yaml
@@ -185,13 +185,25 @@ jobs:
           Get-FileHash MNE-Python-$env:MNE_INSTALLER_VERSION-Windows.exe -Algorithm SHA256 > MNE-Python-$env:MNE_INSTALLER_VERSION-Windows.exe.sha256.txt
           Get-Content MNE-Python-$env:MNE_INSTALLER_VERSION-Windows.exe.sha256.txt
 
-      - name: Run installer (macOS)
-        if: ${{ runner.os == 'macOS' && matrix.arch == 'native' }}
+      - name: Check installer (macOS)
+        if: ${{ runner.os == 'macOS' }}
+        env:
+          ARCH: ${{ matrix.arch }}
         run: |
           echo `pwd`
-          installer -verbose -pkginfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_Intel.pkg
-          installer -verbose -dominfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_Intel.pkg
-          installer -verbose -volinfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_Intel.pkg
+          if [[ ${ARCH} == 'native' ]]; then
+            MACOS_SUFFIX="Intel"
+          else
+            MACOS_SUFFIX="ARM"
+          fi
+          echo "MACOS_SUFFIX=${MACOS_SUFFIX}" >> $GITHUB_ENV
+          installer -verbose -pkginfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_{MACOS_SUFFIX}.pkg
+          installer -verbose -dominfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_{MACOS_SUFFIX}.pkg
+          installer -verbose -volinfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_{MACOS_SUFFIX}.pkg
+
+      - name: Run installer (macOS Intel)
+        if: ${{ runner.os == 'macOS' && matrix.arch == 'native' }}
+        run: |
           sudo installer \
             -verbose \
             -pkg MNE-Python-${MNE_INSTALLER_VERSION}-macOS_Intel.pkg \
@@ -211,7 +223,7 @@ jobs:
         run: |
           .\MNE-Python-%MNE_INSTALLER_VERSION%-Windows.exe /S /InstallationType=JustMe /AddToPath=1
 
-      - name: Export frozen environment definition (macOS)
+      - name: Export frozen environment definition (macOS Intel)
         if: ${{ runner.os == 'macOS' && matrix.arch == 'native' }}
         run: |
           source /Applications/MNE-Python/.mne-python/bin/activate
@@ -234,7 +246,7 @@ jobs:
           mamba list --json > MNE-Python-$env:MNE_INSTALLER_VERSION-Windows.env.json
           Get-Content MNE-Python-$env:MNE_INSTALLER_VERSION-Windows.env.json
 
-      - name: Check installation (macOS)
+      - name: Check installation (macOS Intel)
         if: ${{ runner.os == 'macOS' && matrix.arch == 'native' }}
         run: |
           source /Applications/MNE-Python/.mne-python/bin/activate
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -197,9 +197,9 @@ jobs:
             MACOS_SUFFIX="ARM"
           fi
           echo "MACOS_SUFFIX=${MACOS_SUFFIX}" >> $GITHUB_ENV
-          installer -verbose -pkginfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_{MACOS_SUFFIX}.pkg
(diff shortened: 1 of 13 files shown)


larsoner:
@hoechenberger can you request migration for mayavi? It's needed for mne-kit-gui

hoechenberger:
@larsoner I've made a PR:
https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2910

FSLeyes depends on dcm2niix, btw… 🙈

larsoner:
Okay locally it at least starts building if I exclude the ones from #48 plus `fsleyes`, `mne-kit-gui` (mayavi), and `mne-installer-menus`.

> Ouch, that's going to be a tough one I'm afraid. I will add it to the list of requested migrations, let's see what happens…

Mayavi actually uses very little compiled code so hopefully it won't be too bad. Locally it at least compiled properly, but my env having PyQt6 was a problem when I tried to test it using `mayavi.mlab.test_plot3d()`. But I'm optimistic that it at least compiled fully!

hoechenberger:
> Mayavi actually uses very little compiled code so hopefully it won't be too bad. Locally it at least compiled properly, but my env having PyQt6 was a problem when I tried to test it using `mayavi.mlab.test_plot3d()`. But I'm optimistic that it at least compiled fully!

Nice!

Yes I've also had issues with pip-installed PyQt6 alongside Qt5

larsoner:
@hoechenberger when I get rid of the `# [not arm64]` for `mne-installer-menus`, I get package conflicts locally (where I cannot use the patched constructor). I get the same when I try to crosscompile for ARM on my Intel mac (where I *do* use the patched constructor). So either it needs to be migrated, or for some reason it has conflicts that the `mne` package does not. I don't understand either way really since in the bot issue it's clear it shouldn't be a migration issue, and the package only requires `mne-base` to install. Any ideas there?

I've pushed to CIs to see if they have the same issue. If they do, I'll restore the `# [not arm64]`.

hoechenberger:
@larsoner Only with the last build we [dropped](https://github.com/conda-forge/mne-feedstock/pull/97) the Spyder dependency from `mne-installer-menus`; so I've bumped the build number to use in the installer accordingly in af5197d

hoechenberger:
We're green!

.github/workflows/build.yml
```diff
@@ -173,13 +185,25 @@ jobs:
...
+          ARCH: ${{ matrix.arch }}
         run: |
           echo `pwd`
-          installer -verbose -pkginfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_Intel.pkg
-          installer -verbose -dominfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_Intel.pkg
-          installer -verbose -volinfo -pkg ./MNE-Python-${MNE_INSTALLER_VERSION}-macOS_Intel.pkg
+          if [[ ${ARCH} == 'native' ]]; then
+            MACOS_SUFFIX="Intel"
+          else
+            MACOS_SUFFIX="ARM"
```
hoechenberger, line 197: Since the filenames of the installer packages are "user-facing", I'd suggest we use the suffix "M1" for now, as I assume most Apple Silicon users won't really know what "ARM" is supposed to mean.
```suggestion
            MACOS_SUFFIX="M1"
```

larsoner:
One issue with the `M1` name is that it's likely to become a misnomer soon with M2 chips coming out sometime probably in 2022. But I'm not sure of a better name, and I agree that it's clearer than ARM, so I changed it in the two places it needed to be updated.

hoechenberger:
> One issue with the `M1` name is that it's likely to become a misnomer soon with M2 chips coming out sometime probably in 2022. But I'm not sure of a better name, and I agree that it's clearer than ARM, so I changed it in the two places it needed to be updated.

Yes I thought about this too, but I don't think it will require much mental gymnastics to assume that was made to support M1 is  highly likely to work with M2 as well :)

recipes/mne-python_1.0/construct.yaml
```diff
@@ -25,7 +25,8 @@ default_prefix_all_users: "%ALLUSERSPROFILE%\\mne-python\\1.0.3_0"  # [win]
 
 uninstall_name: MNE-Python ${VERSION} (Python ${PYVERSION})
 
-installer_filename: MNE-Python-1.0.3_0-macOS_Intel.pkg  # [osx]
+installer_filename: MNE-Python-1.0.3_0-macOS_Intel.pkg  # [osx and not arm64]
+installer_filename: MNE-Python-1.0.3_0-macOS_M1.pkg    # [osx and arm64]
```
hoechenberger, line 29:
```suggestion
installer_filename: MNE-Python-1.0.3_0-macOS_M1.pkg     # [osx and arm64]
```

recipes/mne-python_1.0/construct.yaml
```diff
@@ -25,7 +25,8 @@ default_prefix_all_users: "%ALLUSERSPROFILE%\\mne-python\\1.0.3_0"  # [win]
 
 uninstall_name: MNE-Python ${VERSION} (Python ${PYVERSION})
 
-installer_filename: MNE-Python-1.0.3_0-macOS_Intel.pkg  # [osx]
+installer_filename: MNE-Python-1.0.3_0-macOS_Intel.pkg  # [osx and not arm64]
+installer_filename: MNE-Python-1.0.3_0-macOS_ARM.pkg    # [osx and arm64]
```
larsoner, line 29: Errant change from last commit
```suggestion
installer_filename: MNE-Python-1.0.3_0-macOS_M1.pkg    # [osx and arm64]
```

larsoner:
EMD import error should be fixed by https://github.com/conda-forge/emd-feedstock/pull/7

hoechenberger:
@larsoner

We must also adjust the constructor patch. On Apple Silicon, the following lines:

https://github.com/mne-tools/mne-installers/blob/9039eb6e53a66b82d4279a22585bbb6c35215dbc/assets/constructor_macOS.patch#L65-L66

should be replaced with:
```
options.set('hostArchitectures', 'arm64')
```

larsoner:
> So I suppose we should split the patch into a common patch, and two separate ones for each platform that only adjust the line(s) I just mentioned?

Yes that would make sense. Feel free to do it if you want, or I can do it tomorrow

hoechenberger:
Ah no maybe we should keep it for now:

> When you use productbuild to create a distribution pkg on Big Sur (Intel and Apple silicon) both arm64 and x86_64 will be added to the configuration by default.
> 
> But, when you use productbuild on Catalina or earlier, the attribute will be lacking, when means that when someone installs that pkg on an Apple silicon Mac, it will assume it requires Rosetta and prompt for installation.


https://scriptingosx.com/2020/12/platform-support-in-macos-installer-packages-pkg/

hoechenberger:
… buuuut then again, we're only building for Apple Silicon on Big Sur (macOS 11) here anyway, so …

larsoner:
Maybe we should remove the patching for now then, and then try the installers on our M1 machines when the next release is in draft mode. (Or go ahead and build locally on non-M1 now, install on M1.) I think it would be nice if the Intel builds still installed on M1, too, in case people have modules they want to use on top of our installer that are still intel-only.

assets/constructor_macOS.patch
```diff
@@ -62,8 +62,6 @@ diff -Naur constructor-orig/osxpkg.py constructor-patched/osxpkg.py
 +        options.set('customize', 'never')
 +        options.set('customLocation', '/Applications/MNE-Python')
 +        options.set('rootVolumeOnly', 'true')
```
larsoner, line 64: Ohh we still want this stuff so we get rid of the entire patch, just parts of it. Makes sense

hoechenberger: Yes we want to keep that :)

hoechenberger:
@larsoner We need to make a decision regarding those patches :)

Just as a recap:

Currently, we inject a `hostArchitecture` key into the installer which basically says, "this is an Intel application, so if you're not on an Intel machine but on Apple Silicon (AS), run this through Rosetta".

This was needed because we couldn't provide native AS builds in the past.

Now the situation has changed.

## Native AS builds
When we're building natively for AS, we **must** alter `hostArchitecture` thing, otherwise macOS will try to run the native AS binaries through Rosetta (which will fail).

So in the following, we need to change the value to `arm64`:
https://github.com/mne-tools/mne-installers/blob/9039eb6e53a66b82d4279a22585bbb6c35215dbc/assets/constructor_macOS.patch#L65-L66

(I believe simply **removing** this patch and relying on the default might cause trouble too, as I believe then it would default to `x86_64,arm64`, and if somebody tried to run the AS installer on an Intel machine, it would seem to work at first and then fail, as we're not shipping Intel binaries in that installer … but I haven't tested this yet).

So… better be explicit than implicit, and set this to `arm64`.

There is another thing we must revert: The following patch ensured that when running out Intel binaries on AS, `conda` wouldn't try to fetch AS packages during manual `conda` package installations run by the user after the installer had finished:

https://github.com/mne-tools/mne-installers/blob/9039eb6e53a66b82d4279a22585bbb6c35215dbc/assets/post_install_macOS.sh#L33-L35

So we need to remove the `CONDA_SUBDIR` variable (but keep `PYTHONNOUSERSITE`)

## Intel builds

I don't think we need to worry about AS users relying on Intel packages **and** using the installer. Because to me, this seems like a kind of advanced use case, and I doubt that these users would run the installer anyway?

At any rate, I'm open to either approach here: either entirely nix our ability to run on AS, or keep compatibility.

Shall we maybe discuss & resolve this together tomorrow? Otherwise feel free to go ahead here and propose your preferred changes :)

larsoner:
@hoechenberger if I try to create an env on my M1 machine natively with *only* spyder, I now get:
```
  - feature:/osx-arm64::__osx==12.4=0
  - feature:|@/osx-arm64::__osx==12.4=0
  - spyder -> applaunchservices[version='>=0.3.0'] -> __osx
```
I don't really understand this, though, as:
```
$ mamba repoquery search -c conda-forge applaunchservices
...
 Name              Version Build        Channel           
───────────────────────────────────────────────────────────
 applaunchservices 0.3.0   pyhd8ed1ab_2 conda-forge/noarch
 applaunchservices 0.2.1   py_0         conda-forge/noarch
```

hoechenberger:
> Didn't check, so maybe that was it. Given that things are green here, I guess it probably was?

Yep appears so! I can now install Spyder locally on arm64.

hoechenberger:
Nice! I'll leave that to you *cough* :) it's already evening in Paris and I'll sign off soon.

larsoner:
I won't get to it today, working on https://github.com/jaimergp-forge/constructor-feedstock/pull/3

hoechenberger:
@larsoner I pushed one of the required changes (postinstall script), but didn't have time to test locally

We still need to adjust the patchfile

larsoner:
Looks like native is broken now:
```
installer: Error - MNE-Python 1.0.3_0 can’t be installed on this computer.
```

larsoner:
... and then I guess we should try to port these patches upstream first to napari, then maybe help them port them upstream to `conda` proper. WDYT?

hoechenberger:
Thanks for this, @larsoner!

PR #129, opened 2022-06-03
Bump version to 1.0.3_1

hoechenberger:


larsoner:
If we need to help out the conda constructor, it might work to put the docutils and/or jinja pins back in there. I think different libraries have been pinning these in different ways that might be a pain for conda to resolve. You could test locally the resolution speed with and without the pins to see if it helps at all...?

hoechenberger:
@larsoner I don't have time today to test this with different pinnings … but yes you're right, often explicit version specifications help `conda` a lot. Would you like to give it a shot?

hoechenberger:
@larsoner Since `constructor` internally uses `conda` for dependency resolution, maybe we can make it use the `libmamba`  solver?

https://github.com/conda-incubator/conda-libmamba-solver#how-do-i-configure-conda-to-use-the-experimental-solver-permanently

hoechenberger:
@jaimergp Would you consider building a newer `conda` (4.12 or newer) for `napari/label/bundle_tools` on anaconda.org? I wanted to use the `libmamba` solver for dependency resolution in `constructor`, and this requires `conda >=4.12`, but the latest version you provide is 4.12a

jaimergp:
Yes, I need to build 4.13. Possibly next week!

hoechenberger:
That would be amazing! LMK if you need any help building or testing those, @jaimergp!

PR #135, opened 2022-06-08
CI: Add automerge

larsoner:
I just installed https://github.com/apps/prylabs-bulldozer in this repo and added an `automerge` label to this PR. Let's see if it works.


larsoner:
Okay that didn't work, but maybe it needs to be in `main` already. I'll go ahead and merge and we can try again in another PR

PR #136, opened 2022-06-08
ENH: Split and persist CI jobs & artifacts

### LINKED ISSUE #134: ENH: Speed up build+test
Why do we build installers for two versions of each OS (plus M1)? It seems like we should build just on the oldest one for each OS (plus M1), then test on the oldest and the newest version of each OS (plus M1). In other words, it seems like we should have two stages: the first "build" stage on 4 OSes (old ubuntu/mac/win + M1), and a second "test" stage on 7 OSes (old/new x ubuntu/mac/win, plus M1).

larsoner:
Closes #134

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,8 +182,12 @@ jobs:
         shell: bash -el {0}
         run: |
           shopt -s nullglob  # Fail if the following pattern yields no results
+          echo "Finding matches"
           matches=(MNE-Python-*-*.*)
+          echo "Extracting fname"
           installer_fname="${matches[0]}"
+          echo "Found name: ${installer_fname}"
+          echo "Want name:  ${MNE_INSTALLER_NAME}"
           test "$installer_fname" == "$MNE_INSTALLER_NAME"
           hash_fname="${MNE_INSTALLER_NAME}.sha256.txt"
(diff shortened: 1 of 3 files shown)


larsoner:
Ouch, upload takes ~7 minutes. Then all tests are bottlenecked until all 4 of the build jobs finish. I'm becoming skeptical that this will actually help us very much. @hoechenberger WDYT?

larsoner:
(download only takes ~1 minute at least)

hoechenberger:
@larsoner What's your gut feeling on the added-enhancements-to-added-complexity ratio here?

larsoner:
> @larsoner What's your gut feeling on the added-enhancements-to-added-complexity ratio here?

I don't think the complexity is too bad. If anything I think it's nice because it totally isolates the build process from the install process. They happen effectively on different machines now, which is nice.

What I was worried about was additional overhead. I thought this would speed things up but the upload time is so long that I think it's roughly the same as it was before in terms of push-to-green. In terms of overall time, it looks like this PR used ~123min of total CI time on my last push. The last build on `main` used 135 min. So they're pretty comparable in terms of overall utilization.

So overall I'm +1 for merge now

hoechenberger:
Thanks @larsoner!!!!

PR #138, opened 2022-06-12
During automated testing, don't require system-wide installed XCB libraries anymore on Linux

hoechenberger:
Now that https://github.com/conda-forge/qt-main-feedstock/pull/39 has been resolved


hoechenberger:
@larsoner It's actually working!!!!

PR #141, opened 2022-06-17
Use the Qt-based Matplotlib backend by default on macOS

### LINKED ISSUE #140: Set MPLBACKEND env var to enforce using Qt5Agg backend?
Would there be any drawbacks on headless machines, e.g. running through Jupyter Lab? Would Matplotlib just refuse to work there, instead of automatically picking a different backend?

Came up in https://mne.discourse.group/t/semi-interactive-plot-with-matplotlib-backend/5094

hoechenberger:
Fixes #140

### DIFF
--- a/assets/post_install_macOS.sh
+++ b/assets/post_install_macOS.sh
@@ -39,6 +39,9 @@ if [ "${PYTHON_PLATFORM}" == "x86_64" ]; then
     ${DSTROOT}/.mne-python/bin/conda env config vars set CONDA_SUBDIR=osx-64
 fi
 
+logger -p 'install.info' "ℹ️ Configuring Matplotlib to use the Qt backend by default"
+${DSTROOT}/.mne-python/bin/conda env config vars set MPLBACKEND=qtagg
+
 logger -p 'install.info' "ℹ️ Configuring Python to ignore user-installed local packages"
 ${DSTROOT}/.mne-python/bin/conda env config vars set PYTHONNOUSERSITE=1
 
--- a/assets/post_install_macOS.sh
+++ b/assets/post_install_macOS.sh
@@ -39,14 +39,14 @@ if [ "${PYTHON_PLATFORM}" == "x86_64" ]; then
     ${DSTROOT}/.mne-python/bin/conda env config vars set CONDA_SUBDIR=osx-64
 fi
 
-logger -p 'install.info' "ℹ️ Configuring Matplotlib to use the Qt backend by default"
-${DSTROOT}/.mne-python/bin/conda env config vars set MPLBACKEND=qtagg
-
 logger -p 'install.info' "ℹ️ Configuring Python to ignore user-installed local packages"
 ${DSTROOT}/.mne-python/bin/conda env config vars set PYTHONNOUSERSITE=1
 
 logger -p 'install.info' "ℹ️ Disabling mamba package manager banner"
 ${DSTROOT}/.mne-python/bin/conda env config vars set MAMBA_NO_BANNER=1
+
+logger -p 'install.info' "ℹ️ Configuring Matplotlib to use the Qt backend by default"
+${DSTROOT}/.mne-python/bin/conda env config vars set MPLBACKEND=qtagg
 
 logger -p 'install.info' "Fixing permissions of entire conda environment"
 chown -R $USER_FROM_HOMEDIR "${DSTROOT}/.mne-python"
--- a/assets/post_install_macOS.sh
+++ b/assets/post_install_macOS.sh
@@ -46,7 +46,7 @@ logger -p 'install.info' "ℹ️ Disabling mamba package manager banner"
 ${DSTROOT}/.mne-python/bin/conda env config vars set MAMBA_NO_BANNER=1
 
 logger -p 'install.info' "ℹ️ Configuring Matplotlib to use the Qt backend by default"
-${DSTROOT}/.mne-python/bin/conda env config vars set MPLBACKEND=qtagg
+sed -i '.bak' "s/##backend: Agg/backend: qtagg/" ${DSTROOT}/.mne-python/lib/python3.10/site-packages/matplotlib/mpl-data/matplotlibrc
 
 logger -p 'install.info' "Fixing permissions of entire conda environment"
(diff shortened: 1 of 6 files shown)


hoechenberger:
I've changed my mind and I'm defaulting to qtagg on all platforms now. Users in Jupyter will want to run some form of `%matplotlib` magic anyway.

assets/post_install_macOS.sh
```diff
@@ -45,6 +45,9 @@ ${DSTROOT}/.mne-python/bin/conda env config vars set PYTHONNOUSERSITE=1
 logger -p 'install.info' "ℹ️ Disabling mamba package manager banner"
 ${DSTROOT}/.mne-python/bin/conda env config vars set MAMBA_NO_BANNER=1
 
+logger -p 'install.info' "ℹ️ Configuring Matplotlib to use the Qt backend by default"
+${DSTROOT}/.mne-python/bin/conda env config vars set MPLBACKEND=qtagg
```
larsoner, line 49: I'd rather not use an env var because it actually takes precedence over `matplotlibrc`. If possible a better option would be to use the lowest-priority location, which I guess would be (4) here

https://matplotlib.org/stable/tutorials/introductory/customizing.html#the-matplotlibrc-file

hoechenberger: I'd actually **intended** to override the `matplotlibrc`, hence picked the env var 🤔 

Another approach I'd considered is deploying our own MPL config and setting [`MPLCONFIGDIR`](https://matplotlib.org/stable/users/faq/environment_variables_faq.html#envvar-MPLCONFIGDIR), so users would still be able to alter the default via the Matplotlib Python API

WDYT?

larsoner: I would rather not do that. I think a user's matplotlibrc should take precedence if they have one. It allows us to respect the standard order/expectations as much as possible.

Hence 4 in the link above is best I think

larsoner: ... let's say for example for 10 years I've been using `Agg` as my backend by setting it in `~/.matplotlilb/matplotlibrc` (for whatever reason). This would have worked on every platform and in every install automatically. What you currently do here will make it (mysteriously!) stop working. Same thing if we go the MPLCONFIGDIR route (and maybe worse, this approach would ignore/break any *other* config changes people have in there).

Using option (4) above respects it. Hence it's the least invasive (and therefore best) option I think to use (4).

Basically in my mind we *just* want to change the internal matplotlib default from MacOS to Qt. We can't/shouldn't patch MPL's code, so we should instead make the next least invasive / pushy change, which is to ship a `matplotlibrc` in the `matplotlib` installation directory.

hoechenberger: Ok, convinced

hoechenberger:
@larsoner Feel free to merge!

assets/post_install_macOS.sh
```diff
@@ -45,6 +45,9 @@ ${DSTROOT}/.mne-python/bin/conda env config vars set PYTHONNOUSERSITE=1
 logger -p 'install.info' "ℹ️ Disabling mamba package manager banner"
 ${DSTROOT}/.mne-python/bin/conda env config vars set MAMBA_NO_BANNER=1
 
+logger -p 'install.info' "ℹ️ Configuring Matplotlib to use the Qt backend by default"
+sed -i '.bak' "s/##backend: Agg/backend: qtagg/" ${DSTROOT}/.mne-python/lib/python3.10/site-packages/matplotlib/mpl-data/matplotlibrc
```
larsoner, line 49: In 6 months we'll have to remember to change this. Better to add something like the following, adapted from `extract_version.sh`
```
export PYSHORT=$($DSTROOT/.mne-python/bin/python -c "import sys; print('.'.join(map(str, sys.version_info[:2])))")
```

larsoner: ... then use `$PYSHORT` in this line

larsoner: Before you do, it might be worth changing it to 3.9 to verify that CIs complain. If the installer succeeds silently/with no error, then it means this script could get broken halfway through and we'd only know once we saw some step not working in practice later...

hoechenberger: Yep, failing!

assets/post_install_macOS.sh
```diff
@@ -33,26 +33,28 @@ osascript \
 # Use Intel packages if the Python binary is x84_64, i.e. not native Apple Silicon
 # (This also applies to an Intel binary running on Apple Silicon through Rosetta)
 # https://conda-forge.org/docs/user/tipsandtricks.html#installing-apple-intel-packages-on-apple-silicon
-export PYTHON_PLATFORM=`${DSTROOT}/.mne-python/bin/conda run python -c "import platform; print(platform.machine())"`
+DSTBIN=${DSTROOT}/.mne-python/bin
+PYTHON_PLATFORM=$(${DSTBIN}/conda run python -c "import platform; print(platform.machine())")
```
hoechenberger, line 37: Is there a specific reason why you're not using backticks instead of `$(...)`? I find the backticks easier to read

assets/post_install_macOS.sh
```diff
@@ -33,26 +33,28 @@ osascript \
...
     ${DSTROOT}/.mne-python/bin/conda env config vars set CONDA_SUBDIR=osx-64
 fi
 
 logger -p 'install.info' "ℹ️ Configuring Python to ignore user-installed local packages"
-${DSTROOT}/.mne-python/bin/conda env config vars set PYTHONNOUSERSITE=1
+${DSTBIN}/conda env config vars set PYTHONNOUSERSITE=1
 
 logger -p 'install.info' "ℹ️ Disabling mamba package manager banner"
-${DSTROOT}/.mne-python/bin/conda env config vars set MAMBA_NO_BANNER=1
+${DSTBIN}/conda env config vars set MAMBA_NO_BANNER=1
```
hoechenberger, line 48: Nice, we should probably adopt a similar approach in the other postinstall scripts too

assets/post_install_macOS.sh
larsoner: Yes, some BASH linter (in vscode I think?) complained that backticks were bad and that I should use `$()` instead

larsoner:
Thanks @hoechenberger !

PR #144, opened 2022-07-27
Include pycrostates 0.1.2

mscheltienne:
We fixed a reported issue in the packaging: the `.jinja` files were not included in the tarball and the wheel. Oopsie 😅 
It would be nice if this bugfix release could be included in the installers for 1.1. Depends on https://github.com/conda-forge/pycrostates-feedstock/pull/4

One question, what does the operator `~=` do exactly?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,11 +5,13 @@ concurrency:
 
 on:
   push:
-    branches: [ main ]
+    branches:
+      - main
     tags:
     - '*'
   pull_request:
-    branches: [ main ]
+    branches:
+      - main
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
@@ -95,7 +97,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Extract version information
-        shell: ${{ matrix.shell }} -el {0}
+        shell: '${{ matrix.shell }} -el {0}'
         env:
           MNE_CROSSCOMPILE_ARCH: ${{ matrix.arch }}
         run: |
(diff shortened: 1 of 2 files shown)


hoechenberger:
> One question, what does the operator `~=` do exactly?

It's like `>=`, but restricted to the last section of the version number. For example, `~= 1.2.3` basically means `1.2.3 <= X < 1.3.0`

`~= 1.2` basically means `1.2 <= X < 2.0`

It's called a "Compatible release clause"

See https://peps.python.org/pep-0440/#version-specifiers

mscheltienne:
Oh! So `~=0.1.1` will choose `0.1.2` if available.
I could not figure out the proper keyword in Google to get an answer, thanks!

hoechenberger:
> Oh! So `~=0.1.1` will choose `0.1.2` if available.

Yep.
But let's bump it just to be on the safe side.

hoechenberger:
@larsoner 
It appears we're having an MSYS2-related issue on the Windows CI … no idea where this is coming from

hoechenberger:
@larsoner It's interesting that apparently bash works flawlessly before we install Micromamba, and in the step directly after installation, bash stops working.

hoechenberger:
The bash executable changes after micromamba install:

Screen Shot 2022-07-27 at 22 09 38

hoechenberger:
It's still trying to use Git bash, ugh…

hoechenberger:
Seems we cannot use matrix to select the shell either. Oh man this is just so frustrating.

mscheltienne:
:tada: :tada:

hoechenberger:
Ok this seems to work

@larsoner I've switched to using MSYS2 on Windows so we have more explicit control over which shell gets used. This fixed the issues we saw initially on CI with this PR. Only annoying thing is that conda environment activation doesn't happen automatically, but I suppose we can try to fix that later.

PR #148, opened 2022-07-29
Update numerous bundled packages and add automated check for outdated packages

hoechenberger:
@larsoner This is WIP work to automatically find out which packages are potentially outdated.

My idea is to add this to a CI job to run weekly or so.

Please feel free to take over if you have time. It's already sort of useful, I suppose:

```
mne is up to date
mne-installer-menus not found on PyPI
openblas not found on PyPI
mne-qt-browser is outdated
mne-bids is up to date
mne-connectivity is up to date
mne-faster is up to date
mne-nirs is outdated
mne-realtime is outdated
mne-features is up to date
mne-rsa is outdated
mne-ari is outdated
mne-kit-gui is up to date
mne-icalabel is outdated
autoreject is up to date
pyprep is outdated
python not found on PyPI
mne-microstates is up to date
pycrostates is up to date

3 packages not found on PyPI:
[Package(name='mne-installer-menus', version='1.1dev0'), Package(name='openblas', version='0.3.20'), Package(name='python', version='3.10.0')]

7 packages are outdated:
[Package(name='mne-qt-browser', version='0.3.0'), Package(name='mne-nirs', version='0.2.0'), Package(name='mne-realtime', version='0.1.0'), Package(name='mne-rsa', version='0.6.0'), Package(name='mne-ari', version='0.1.0'), Package(name='mne-icalabel', version='0.1.0'), Package(name='pyprep', version='0.4.0')]
```

PR #149, opened 2022-08-03
MAINT: Bump to 1.1

larsoner:
This won't work until @drammock cuts a 1.1.0 release on Github, conda-forge updates the feedstock, and then the CDNs update, but in the meantime I wanted to set this up so it's ready to go (restart CIs then be green hopefully!).

@hoechenberger feel free to review to make sure this is reasonable. I changed some `=0.3.1` to `~=0.3.1` since it seems like we should be okay with newer patch releases being installed, right?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: macos-10.15
+          - os: macos-11
             arch: x86_64
             shell: bash
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -240,7 +240,7 @@ jobs:
     needs: build
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


hoechenberger:
@larsoner With #148 I changed the way we specify versions, that is, I think it makes sense to use explicit pins to one specific version (as in contrast to pinning to all compatible versions) 

This ensures that all version specifications are now very explicit. Thanks to that small script I've added, it's now simple to check whether anything is outdated. 

I would therefore prefer if you could roll back the changes to pinning you're making in this PR, unless you have some convincing ideas as to why it's the better way?

larsoner:
> unless you have some convincing ideas as to why it's the better way?

Generally patch releases fix bugs, so we're potentially missing out on bugfixes. But being able to check versions is nice so I'll roll it back

hoechenberger:
Btw I just realized my script doesn't check the version of Python

PR #150, opened 2022-08-24
MAINT: Bump to 1.1.1

larsoner:
Will require https://github.com/conda-forge/mne-feedstock/pull/106 and CDN update but let's get the ball rolling

### DIFF
--- a/.github/workflows/outdated.yml
+++ b/.github/workflows/outdated.yml
@@ -14,9 +14,6 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v3
-    - uses: conda-incubator/setup-miniconda@v2
-      with:
-        activate-environment: "test"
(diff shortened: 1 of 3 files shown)


hoechenberger:
@larsoner Thanks for this! Can you please also run [test_outdated.py](https://github.com/mne-tools/mne-installers/blob/main/tests/test_outdated.py) to see if we can bump a few other package versions, too? For example, I believe there was a new MNE-IClabel release just recently

larsoner:
> Can you please also run [test_outdated.py](https://github.com/mne-tools/mne-installers/blob/main/tests/test_outdated.py)

I don't like such manual steps so I added it to CI :)

larsoner:
Correctly detects that I (intentionally) left out `mne-icalabel=0.4`:

https://github.com/mne-tools/mne-installers/runs/7999062080?check_suite_focus=true

PR #151, opened 2022-10-03
MAINT: Add 1.2 installer

larsoner:
In prep for 1.2 release, let's try making the installers. This still installs MNE 1.1.1 so it should fail. But I've updated the deps based on `tests/test_outdated.py` at least and tried to mark what I think will also need to be updated.


hoechenberger:
Tests are failing because latest released autoreject doesn't support MNE 1.2 (but fix exists in autoreject's master branch)

hoechenberger:
Nice 👍 we're green

drammock:
@larsoner I'm signing off for the night; the only thing I didn't get to (besides email/discourse/discord announcements) is this:

> Update documentation links to installers. Change button links in doc/install/installers.rst to point to the new installer files. Make sure to update links for all 3 platforms. Push this change to both maint/X.Y and main, with commit message [circle deploy].

If you're up late, feel free to handle that one.

PR #153, opened 2022-10-11
bump to v1.2

drammock:
first time doing this so probably I did some things wrong; please look carefully @hoechenberger and @larsoner


larsoner:
Can you push to #151 instead? You should be able to comment out the dev stuff and uncomment the stable lines in the recipe

drammock:
> Can you push to #151 instead? You should be able to comment out the dev stuff and uncomment the stable lines in the recipe

permission denied

larsoner:
Made you admin and also pushed. Will cut a release here once it comes back green, then update MNE stable and dev

PR #158, opened 2022-11-29
Include seaborn, plotly, and the MNE-BIDS-Pipeline

### LINKED ISSUE #156: Pin ipywidgets to <8
Otherwise, tqdm may not work properly in VS Code:
https://github.com/microsoft/vscode-jupyter/issues/8552#issuecomment-1263940999

### LINKED ISSUE #157: Include seaborn
What do you think about adding seaborn to the constructor?

mscheltienne:
Closes #157
Closes #156

Couple of points/questions:
- `pqdm` is already listed (not that I know what it is)
- Do you want to explicitly list `matplotlib`?
- `plotly` is not distributed on `conda-forge`, do you want to add the `plotly` channel?


larsoner:
> plotly is not distributed on conda-forge, do you want to add the plotly channel?

I see https://github.com/conda-forge/plotly-feedstock and locally:
```
$ conda create -n plotly -c conda-forge plotly
Collecting package metadata (current_repodata.json): done
Solving environment: done

... # a bunch of normal install stuff
```
I'll push a commit to add it here and we'll see if it works (it should).

The Windows error is not good, though :unamused: It's probably not related to your changes, though, so we could merge once the others are green.

mscheltienne:
I've never used `plotly`, but the recommended install is via `conda install -c plotly plotly=5.11.0` even if there is a feedstock on conda-forge. Any idea why?

That windows error had an interesting exit code 😞

larsoner:
> but the recommended install is via conda install -c plotly plotly=5.11.0 even if there is a feedstock on conda-forge. Any idea why?

No I have never looked into this. It would be a good question for some support forum for them, or even https://github.com/conda-forge/plotly-feedstock/issues

hoechenberger:
I use plotly from conda-forge daily. This is in a Linux container running in Docker.

hoechenberger:
Maybe we can address #156 here too, it might help the solver...

recipes/mne-python_1.2/construct.yaml
```diff
@@ -129,6 +129,12 @@ specs:
   # NeuroSpin needs the following
   - questionary
   - pqdm
+  # Viz
+  - seaborn
+  - plotly
+  # Temporary
+  # for Vscode compatibility: https://github.com/microsoft/vscode-jupyter/issues/8552#issuecomment-1263940999
+  - ipywidgets <=7.7.2
```
hoechenberger, line 137:
```suggestion
  - ipywidgets <8
```

hoechenberger:
Otherwise LGTM!

tests/test_outdated.py
```diff
@@ -28,6 +28,11 @@ class Package:
     version_spec: str | None
     version_conda_forge: str | None = None
 
+
+allowed_outdated: set[str] = {
```
hoechenberger, line 32: Why a set though and not just a list?

mscheltienne: Uniqueness isn't a bad idea?

hoechenberger: It was really meant as an open question :) thanks for the explanation, makes sense to me!

hoechenberger (approved):

tests/test_outdated.py
mscheltienne: And that's just my guess :) Maybe there is a different reason.

hoechenberger:
Oh wait, we must bump the installer version number before merging!

mscheltienne:
Can you push the corresponding changes? I'm not sure I know where this version number is.

hoechenberger:
I somehow couldn't push to your branch, so I created a PR:
https://github.dev/mscheltienne/mne-installers/pull/1

hoechenberger:
@larsoner Windows installations are failing without any helpful information :(

larsoner:
No idea what's going on there :(

hoechenberger:
We should try if `main` still builds

Or maybe someone with a windows machine can try to download the artifact and run the installer locally? Maybe it's some resource Limitation on the CI system? (Out of memory or so)

mscheltienne:
So.. I never used the installers before, but I'm going to say that there is an issue.
I downloaded the artifact for windows and ran it, and it kind of closed a bit early than I would expect (by the end of the main progress bar, without any 'end' screen).
I can't find any shortcuts on my Desktop or in my Start Menu, and it's a pain to remove from my computer (I tried it twice).

Checksum was correct, and I have no idea on how to debug that.

hoechenberger:
@mscheltienne thanks for trying! So is this with the stable installers?

mscheltienne:
Exactly what I'm doing now. Yes, I tried the artifact from the GH action in this PR.

mscheltienne:
With the stable, I did get an error at the end `Error: running C:\ProgramData\....\post_install.bat failed`, BUT, I am getting the environment and the shortcuts in the Start Menu. It took significantly longer to install (with the progress bar almost full, so I guess the artifact version crashes early in the installation process, despite the progress bar being also almost 'full') and it's working fine. Both prompt and Spyder are working.

hoechenberger:
@jaimergp Could you maybe help us here?

We're still using the `napari` fork of `constructor`:

https://github.com/mne-tools/mne-installers/blob/646ddd40aa3af6e248ae4c2af29d8fbe6c84cb1c/environment.yml#L3-L11

But since last week or so, there are problems with Windows installers. While building the installers does work, running them doesn't seem to work. Our respective CI jobs simply die without any helpful logging output, see e.g. here:

https://github.com/mne-tools/mne-installers/actions/runs/3639973746/jobs/6144616527

```shell
Run .\%MNE_INSTALLER_NAME% /S /InstallationType=JustMe /AddToPath=1
  .\%MNE_INSTALLER_NAME% /S /InstallationType=JustMe /AddToPath=1
  shell: C:\Windows\system32\cmd.EXE /D /E:ON /V:OFF /S /C "CALL "{0}""
  env:
    MNE_INSTALLER_ARTIFACT_ID: MNE-Python-Windows-x86_64
    MNE_INSTALLER_NAME: MNE-Python-1.2.3_0-Windows.exe
    MNE_INSTALLER_VERSION: 1.2.3_0
Error: Process completed with exit code -1073741819.
```

Is there any way to get more debugging output somehow?

Thanks in advance for any suggestions you might have!!

jaimergp:
Maybe you are running into the [NSIS 3.08 incompatibility issue](https://github.com/conda/constructor/issues/526)? I am just done with some updates for napari so maybe you can use that now? The new packages are in `napari/label/bundle_tools_2`. There are some minor changes, but it should be essentially the same (only `extra_files` comes to mind).

mscheltienne:
That's a meaningful error at least:

```
  File "C:\Users\runneradmin\mne-python\1.2.3_0\Scripts\mamba-script.py", line 6, in <module>
    from mamba.mamba import main
  File "C:\Users\runneradmin\mne-python\1.2.3_0\lib\site-packages\mamba\mamba.py", line 49, in <module>
    import libmambapy as api
  File "C:\Users\runneradmin\mne-python\1.2.3_0\lib\site-packages\libmambapy\__init__.py", line 7, in <module>
    raise e
  File "C:\Users\runneradmin\mne-python\1.2.3_0\lib\site-packages\libmambapy\__init__.py", line 4, in <module>
    from libmambapy.bindings import *  # noqa: F401,F403
ImportError: DLL load failed while importing bindings: The specified module could not be found.
```

hoechenberger:
Thanks for your suggestions, @jaimergp! Do you have any idea what might be the cause of the error message in the comment above (DLL import problem)? Could it be as simple as a missing runtime dependency?

recipes/mne-python_1.2/construct.yaml
```diff
@@ -76,18 +77,17 @@ specs:
   - autoreject =0.4.0
   - pyprep =0.4.2
   # Python
-  - python =3.10.6
+  - python =3.10.8
   - pip
   - conda
-  - openblas =0.3.21
   - mamba
```
hoechenberger, line 83:
```suggestion
  #  - mamba
```

jaimergp:
Maybe https://github.com/conda/conda/issues/12161? There are some issues with DLLs on Windows in the last release, so keep an eye on the conda issue tracker.

hoechenberger:
Indeed, removing mamba from our dependency spec list fixed the problem (CI still fails because later steps expect mamba to be present)

larsoner:
@hoechenberger I'm going to try keeping `mamba` and just changing our `mamba` calls to `conda` in our tests. Then I can test the installer on Windows and see if `mamba` actually works there or not. That will tell us if end users should have it or not -- ideally we'd keep it and this will just be a CI/infrastructure hangup for us rather than an end-user usability bug

hoechenberger:
@larsoner Care to investigate? It might have to do with the way (in which shell) we run the installer on Windows

larsoner:
> It might have to do with the way (in which shell) we run the installer on Windows

That's what I'm assuming/hoping, that way we can just work around it with `conda ...` here, knowing it works for end-users (assuming my test of `mamba` on the artifact actually works!). That way we can merge this and release a 1.2.3 installer, and look into our `conda` CI setup separately. It's always seemed a bit odd that we have to source the environment script every time just on Windows, but who knows 🤷

Since this sort of investigation can be a bit open-ended, I'd rather get this green, test it locally, and merge, rather than investigate more in this PR

hoechenberger:
But how would we actually **get** this one green? 🤔

larsoner:
It already is. It was failing because we *used* mamba in our commands, not (just) because it was installed. So I've already changed those to use conda instead.

hoechenberger:
Oh wow, I didn't realize! Ok cool!!!

hoechenberger:
@mscheltienne Do you dare to try the windows installer again?

mscheltienne:
Sure, but I won't have access to my Windows machine until Monday.. I'll test it then if this PR is still open :)

larsoner:
I just booted into windows, downloaded the installer, ran it, and verified that `mne sys_info` worked as well as `mamba list` from the MNE terminal, so in it goes. Good teamwork @hoechenberger @mscheltienne !

I'll cut a 1.2.3 installer release next. We're releasing 1.3 at the end of next week but it'll be good to make sure all the release mechanics are still working

PR #161, opened 2022-12-09
MAINT: Use bash for all shells and test properly

larsoner:
- [x] Add 1.3 build (still installing 1.2 until we actually release 1.3)
- [x] ~~Add OpenMEEG~~ OpenMEEG will be a subsequent PR
- [x] Try `bash` as shell on all systems (instead of `msys2`)
- [x] See if the `mamba` problem is fixed
- [x] Reenable non-Windows builds and all packages
- [x] Merge https://github.com/mne-tools/mne-installers/pull/162
- [x] Remove 1.3 build
- [ ] Merge when green
- [ ] ~~Install MNE-Python 1.3 once released~~
- [ ] ~~Merge~~
- [ ] ~~Cut 1.3.0 installer release~~


.github/workflows/build.yml
```diff
@@ -347,171 +335,113 @@ jobs:
...
-          test `grep "Terminal=true"  MNE-Python*.desktop | wc -l` -ge 1
-          test `grep "Terminal=false" MNE-Python*.desktop | wc -l` -ge 1
-
-          # Display their contents
-          for f in MNE-Python*.desktop; do echo "📂 $f:"; cat "$f"; echo; done
-          popd
+          if [[ "${{ runner.os }}" == "macOS" ]]; then
+            python -c "import os; x = os.getenv('CONDA_SUBDIR'); assert x == 'osx-64', f'CONDA_SUBDIR ({repr(x)}) != osx-64'" || exit 1
+          fi
+          # TODO: broken on Windows!
```
larsoner, line 422: These tests are present but not working on `main`:
```
  File "<string>", line 1, in <module>
AssertionError
  File "<string>", line 1, in <module>
AssertionError
Running MNE's sys_info
```
because there is no `-e` there. This PR fixes the `-e`, but these have to be inside a conditional until we fix them on Windows (assuming they are actually meant to be fixed there)

PR #163, opened 2022-12-13
Include MNELAB

### LINKED ISSUE #82: Add MNELAB (requires Qt 6)
As soon as we have Qt6 on `conda-forge`

To keep track: https://github.com/conda-forge/qt-main-feedstock/pull/14

hoechenberger:
Closes #82

Just want to see if it works


hoechenberger:
The ARM build is expected to fail, as we're still waiting for the migration to complete on conda-forge

larsoner:
All green except ARM as expected, great!

hoechenberger:
If you have time, could you download and test the artifacts?

Also I think we need some sort of CI test for MNELAB

Mind you, I believe we now have Qt5 and Qt6 installed in parallel! Who knows what kind d of interactions might occur 😱

larsoner:
... actually I'm not sure if mayavi is compatible. We'd need to check for mne-kit-gui :(

hoechenberger:
I don't think spyder supports
Qt6 either

hoechenberger:
I remember the conda-forge folks spending quite some time to ensure isolation between Qt5 and Qt6, so at least in theory things should work. But who knows how well downstream packages can handle the simultaneous presence of both

hoechenberger:
The ARM build is now on conda-forge, restarting CI.

hoechenberger:
Ouch, dependency resolution error 😳 this is going to be a pain to pin down 😑

larsoner:
Looks like it's actually a conflict with `vtk =9.2.2`. With even just:
```
name: test
channels:
  - conda-forge
dependencies:
  - mnelab =0.8.5
  - python =3.10.8
  - vtk =9.2.2
```
I get:
```
...
UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions

Package python conflicts for:
vtk=9.2.2 -> loguru -> python[version='3.10.*|3.11.*|>=3.5|>=3.6|3.8.*|3.9.*']
mnelab=0.8.5 -> python[version='>=3.10,<3.11.0a0|>=3.10,<3.11.0a0|>=3.8,<3.9.0a0|>=3.8,<3.9.0a0|>=3.9,<3.10.0a0',build=*_cpython]
vtk=9.2.2 -> python[version='>=3.10,<3.11.0a0|>=3.10,<3.11.0a0|>=3.11,<3.12.0a0|>=3.8,<3.9.0a0|>=3.8,<3.9.0a0|>=3.9,<3.10.0a0|>=3.9,<3.10.0a0',build=*_cpython]
python=3.10.8
mnelab=0.8.5 -> matplotlib-base[version='>=3.6.2'] -> python[version='3.10.*|3.8.*|3.9.*|>=3.11,<3.12.0a0|>=3.7|>=3.6|>=3.5',build=*_cpython]

Package expat conflicts for:
vtk=9.2.2 -> expat[version='>=2.4.9,<3.0a0|>=2.5.0,<3.0a0']
vtk=9.2.2 -> python[version='>=3.9,<3.10.0a0'] -> expat[version='>=2.4.1,<3.0a0']

Package setuptools conflicts for:
python=3.10.8 -> pip -> setuptools
mnelab=0.8.5 -> mne-base[version='>=1.0.0'] -> setuptools

Package freetype conflicts for:
vtk=9.2.2 -> ffmpeg[version='>=4.4.2,<5.0a0'] -> freetype[version='>=2.10.4,<3.0a0']
vtk=9.2.2 -> freetype[version='>=2.12.1,<3.0a0']

Package libcblas conflicts for:
mnelab=0.8.5 -> numpy[version='>=1.20.0'] -> libcblas[version='>=3.9.0,<4.0a0']
vtk=9.2.2 -> numpy -> libcblas[version='>=3.9.0,<4.0a0']
```
Rolling back to vtk 9.1 doesn't help because then there are conflicts with openmeeg :(

cbrnr:
Is there anything MNELAB can do to help solve this conflict? I'm not pinning any dependencies (only `>=`) because MNELAB should work always work with the latest versions. Python versions 3.8–3.11 should be supported (I'm aiming for the last three versions, right now these are even four).

hoechenberger:
@larsoner I suspect it has to do with me pinning matplotlib to the latest version... But 3.6.2 is required to work around bugs in PySide 6.4 😩

https://github.com/cbrnr/mnelab/issues/362

hoechenberger:
> @larsoner I suspect it has to do with me pinning matplotlib to the latest version... But 3.6.2 is required to work around bugs in PySide 6.4 😩
> 
> [cbrnr/mnelab#362](https://github.com/cbrnr/mnelab/issues/362)

This assumption was incorrect. I can reproduce the issue by simply requesting vtk and pyside6.

```
❯ mamba create -n test pyside6 vtk=9.2                                                                     ─╯
[...]
Encountered problems while solving:
  - package pyside6-6.4.0-py310h6bbb1e2_0 requires libclang >=13.0.1,<14.0a0, but none of the providers can be installed
```
We should get in touch with the VTK feedstock maintainers.

hoechenberger:
I reported it to the VTK and PySide6 feedstock maintainers:

https://github.com/conda-forge/vtk-feedstock/issues/269
https://github.com/conda-forge/pyside2-feedstock/issues/159

larsoner:
Great @hoechenberger ! Now we wait ...

hoechenberger:
@larsoner Could be a long time…

https://github.com/conda-forge/pyside2-feedstock/issues/159#issuecomment-1350932070

hoechenberger:
I think we should simply build the M1 installers without MNELAB, as the ETA of the OpenSSL migration, which seems to be required to resolve the issue, is unclear (and probably a few months away)

larsoner:
Works for me, feel free to take out of draft mode and mark for merge-when-green @hoechenberger

hoechenberger:
@larsoner Are we sure the Windows installer problems @mscheltienne observed recently are gone now, too? I cannot test, unfortunately

larsoner:
I tested locally on Windows last week and things were fine

hoechenberger:
Great! Enabling automerge.

hoechenberger:
I cannot run MNELAB on M1:

```
❯ mamba create -n test mnelab
❯ mamba activate test
❯ mnelab
2022-12-14 13:40:27.154 python[43193:310391] ApplePersistenceIgnoreState: Existing state will not be touched. New state will be written to /var/folders/2k/nkw3dkx97ssdv08xmdgfnjkc0000gn/T/com.continuum.python.savedState
  File "/Users/hoechenberger/mambaforge/envs/test/bin/mnelab", line 11, in <module>
    sys.exit(main())
  File "/Users/hoechenberger/mambaforge/envs/test/lib/python3.10/site-packages/mnelab/__init__.py", line 42, in main
    model.view = MainWindow(model)
  File "/Users/hoechenberger/mambaforge/envs/test/lib/python3.10/site-packages/mnelab/mainwindow.py", line 285, in __init__
    QKeySequence(Qt.CTRL | Qt.Key_Y),
TypeError: unsupported operand type(s) for |: 'Modifier' and 'Key'
```

hoechenberger:
@cbrnr Here's a fix:
https://github.com/spesmilo/electrum/commit/1478d9d87805b3c03b261908aa125beb711e46ba

hoechenberger:
@larsoner Seems we've got more serious problems with the M1 build than I'd thought … it won't even resolve the deps when MNELAB is not included, it seems?

hoechenberger:
Or is the `arm64` selector perhaps not working 🤔

mscheltienne:
> I don't think spyder supports Qt6 either

It definitely does not. https://github.com/spyder-ide/spyder/blob/e5334d1c1e0b566147739b766d95127d145d0e3d/spyder/requirements.py#L37

hoechenberger:
> Or is the `arm64` selector perhaps not working 🤔

I'm still not sure, but according to the build logs, we're trying to include pyside6 on ARM, which is bound to fail. Not sure where this is coming from though.

recipes/mne-python_1.2/construct.yaml
```diff
@@ -74,6 +74,8 @@ specs:
   - mne-ari =0.1.1
   - mne-kit-gui =1.0.1
   - mne-icalabel =0.4  # [not win]
+  - mnelab =0.8.5   # [not arm64]  # https://github.com/conda-forge/pyside2-feedstock/issues/159
+  - pyside6 =6.4.1  # [not arm64]  # For MNELAB. Without this pinning, 6.4.0 might get installed (why?!)
```
larsoner, line 78:
```suggestion
  # https://github.com/conda-forge/pyside2-feedstock/issues/159
  - mnelab =0.8.5   # [not arm64]
  # For MNELAB. Without this pinning, 6.4.0 might get installed (why?!)
  - pyside6 =6.4.1  # [not arm64]
```

hoechenberger: That was the issue? D'oh! 😅 thanks for figuring it out!!!

hoechenberger:
Whoops. What's up now 🤯

larsoner:
Looks like a conflict between Qt5 and Qt6

hoechenberger:
Do we still need Qt5? If we drop Spyder, ...

mscheltienne:
Would it be possible to maintain at least `spyder-kernels`? At least this way the environment can be quickly linked to an outside Spyder instance.

hoechenberger:
There are tutorials out there where they embed Qt5 Matplotlib plots in a Qt6 application, so they *should* theoretically be able to live side-by-side…

larsoner:
We have a test that PyQt5 works for this repo. We should add an equivalent PySide6 test. It should fail based on what you write above I think...

hoechenberger:
> @larsoner

> **but**
> 
> ```shell
> ❯ mamba create -n test qt6-main
> ❯ ls -lah ~/mambaforge/envs/test/plugins/platforms
> ls: mambaforge/envs/test/plugins/platforms: No such file or directory
> ```

Oh:

```shell
❯ ls -lah ~/mambaforge/envs/test/lib/qt6/plugins/platforms
total 2432
drwxr-xr-x   5 hoechenberger  staff   160B Dec 14 19:33 .
drwxr-xr-x  20 hoechenberger  staff   640B Dec 14 19:33 ..
-rwxrwxr-x   2 hoechenberger  staff   987K Oct 19 11:19 libqcocoa.dylib
-rwxrwxr-x   2 hoechenberger  staff    73K Oct 19 11:19 libqminimal.dylib
-rwxrwxr-x   2 hoechenberger  staff   150K Oct 19 11:19 libqoffscreen.dylib
```

So the plugins have moved. Maybe we need to set an environment variable or something…

hoechenberger:
```shell
❯ mamba create -n test pyside6 pyqt matplotlib-base
❯ mamba run -n test python -c "import matplotlib; matplotlib.use('Qt5Agg')"
```
This works…

hoechenberger:
Oh! Managed to reproduce.

```shell
❯ mamba create -n test pyside6 pyqt matplotlib-base
❯ mamba run -n test python -c "import matplotlib; import matplotlib.pyplot as plt; matplotlib.use('QtAgg'); plt.subplots()"
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/platforms/libqwebgl.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/platforms/libqoffscreen.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/platforms/libqminimal.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/platforms/libqcocoa.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
qt.qpa.plugin: Could not find the Qt platform plugin "cocoa" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

/var/folders/2k/nkw3dkx97ssdv08xmdgfnjkc0000gn/T/tmp92fve324: line 3: 89284 Abort trap: 6           python -c 'import matplotlib; import matplotlib.pyplot as plt; matplotlib.use('"'"'QtAgg'"'"'); plt.subplots()'

ERROR conda.cli.main_run:execute(49): `conda run python -c import matplotlib; import matplotlib.pyplot as plt; matplotlib.use('QtAgg'); plt.subplots()` failed. (See above for error)
```

But this works:
```shell
❯ mamba run -n test python -c "import matplotlib; import matplotlib.pyplot as plt; matplotlib.use('Qt5Agg'); plt.subplots()"
```

larsoner:
Pushed a commit with the suggested test I referred to before (check if pyside6 works at all by itself first)

larsoner:
It appears that even just trying the basic command fails, even on Ubuntu:
```
https://github.com/mne-tools/mne-installers/actions/runs/3698421716/jobs/6265118735#step:13:1945
```
This is likely an untested case at the conda-forge end, and/or they don't support simultaneous Qt5/Qt6 installs, not sure...

hoechenberger:
> and/or they don't support simultaneous Qt5/Qt6 installs

If I'm not mistaken, I remember the person contributing the Qt6 packages spent a **significant** amount of time figuring out how to install both packages in parallel / how to isolate them. But I'm not sure if parallel install was actually extensively **tested**...

hoechenberger:
@larsoner 

And apparently, [`QT_PLUGIN_PATH`](https://doc.qt.io/qt-6/deployment-plugins.html) is being ignored…

```shell
❯ QT_PLUGIN_PATH=/Users/hoechenberger/mambaforge/envs/test/lib/qt6/plugins python -c "from PySide6.QtWidgets import QApplication, QWidget; app = QApplication([])"
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/platforms/libqwebgl.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/platforms/libqoffscreen.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/platforms/libqminimal.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/platforms/libqcocoa.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/platformthemes/libqxdgdesktopportal.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
qt.core.plugin.loader: In /Users/hoechenberger/mambaforge/envs/test/plugins/styles/libqmacstyle.dylib:
  Plugin uses incompatible Qt library (5.15.0) [release]
```

hoechenberger:
Setting `QT_QPA_PLATFORM_PLUGIN_PATH` doesn't seem to help either.

hoechenberger:
@larsoner I'm hoping to generate more and more specific debugging output via [b8d5ecc](https://github.com/mne-tools/mne-installers/pull/163/commits/b8d5eccc1a05a9fbf39436dc304a8a2eb8183cea)

assets/post_install_macOS.sh
```diff
@@ -44,12 +44,12 @@ fi
 logger -p 'install.info' "ℹ️ Configuring Python to ignore user-installed local packages"
 ${DSTBIN}/conda env config vars set PYTHONNOUSERSITE=1
 
+echo "ℹ️ Configuring Matplotlib to use the Qt5 backend by default."
```
hoechenberger, line 47:
```suggestion
logger -p 'install.info' "ℹ️ Configuring Matplotlib to use the Qt5 backend by default."
```

cbrnr:
> I cannot run MNELAB on M1:
> 
> ```
> ❯ mamba create -n test mnelab
> ❯ mamba activate test
> ❯ mnelab
> ...
> TypeError: unsupported operand type(s) for |: 'Modifier' and 'Key'
> ```
> 
> @cbrnr Is this a bug in MNELAB? [pyinstaller/pyinstaller#7249 (comment)](https://github.com/pyinstaller/pyinstaller/issues/7249#issuecomment-1316888347)

No, this is a breaking change from PySide6 < 6.4 vs. PySide6 ≥ 6.4. I adapted my code so that it works with the new types, but forgot to bump the dependency. I will do that right now!

Which version of PySide6 were you running when this problem happened?

hoechenberger:
This was with PySide 6.4.1

cbrnr:
> This was with PySide 6.4.1

Are you sure? The code works with PySide6 ≥ 6.4, I just tested it with 6.4.0 and 6.4.1. The error message indicates that you were running PySide6 < 6.4, before they changed the types of some constants.

hoechenberger:
> Are you sure?

Yes:

```shell
❯ mamba create -y -n test pyside6=6.4.1 mnelab=0.8.5
❯ mamba run -n test mnelab
2022-12-15 11:10:34.021 python[21044:872306] ApplePersistenceIgnoreState: Existing state will not be touched. New state will be written to /var/folders/2k/nkw3dkx97ssdv08xmdgfnjkc0000gn/T/com.continuum.python.savedState
  File "/Users/hoechenberger/mambaforge/envs/test/bin/mnelab", line 11, in <module>
    sys.exit(main())
  File "/Users/hoechenberger/mambaforge/envs/test/lib/python3.10/site-packages/mnelab/__init__.py", line 42, in main
    model.view = MainWindow(model)
  File "/Users/hoechenberger/mambaforge/envs/test/lib/python3.10/site-packages/mnelab/mainwindow.py", line 285, in __init__
    QKeySequence(Qt.CTRL | Qt.Key_Y),
TypeError: unsupported operand type(s) for |: 'Modifier' and 'Key'


ERROR conda.cli.main_run:execute(49): `conda run mnelab` failed. (See above for error)

❯ mamba run -n test python -c "import PySide6; print(PySide6.__version__)"
6.4.1
```

hoechenberger:
@cbrnr The problem does **not** occur with PySide6 6.4.1 installed via `pip`. Uuuughh....

hoechenberger:
I've opened an issue: https://github.com/conda-forge/pyside2-feedstock/issues/164

hoechenberger:
@larsoner I believe we need to wait until https://github.com/conda-forge/pyside2-feedstock/pull/165 has been closed,otherwise ARM builds will continue to fail

recipes/mne-python_1.3/construct.yaml
```diff
@@ -74,6 +74,8 @@ specs:
   - mne-ari =0.1.2
   - mne-kit-gui =1.0.1
   - mne-icalabel =0.4  # [not win]
+  - mnelab =0.8.5
+  - pyside6 =6.4.1
```
hoechenberger, line 78: We may be able to remove the version pinning (or that entire dependency altogether, as it should get installed implicitly when requesting mnelab)

hoechenberger, line 78:
```suggestion
```

larsoner:
Qt5/Qt6 simultaneous install problems remain. Might have to wait until we can get PySide6 only. Seems safest anyway

hoechenberger:
Now with https://github.com/conda-forge/pyside2-feedstock/issues/164 resolved, MNELAB works for me on my M1 Mac:

```shell
❯ mamba create -n mnelab mnelab "pyside6=*=*_3"
❯ mamba run -n mnelab mnelab
```
Screenshot 2023-01-10 at 14 52 40

cbrnr:
This is great! Finally! Thanks for your persistence everyone!

hoechenberger:
Don't celebrate too early though, we still need to address some issues before we can ship MNELAB with the installers. But we're one giant step closer now!!

larsoner:
@hoechenberger any idea if they plan to build PyQt6? For me PySide6 is unusable in Linux because of [this](https://github.com/matplotlib/matplotlib/issues/15359), at least when using `pip` (i.e., `plt.ion()` does not work because PySide6 does not set up the PyOS_InputHook like PyQt6 does)

cbrnr:
Using IPython is not an option for you? If not, do you think a feature request would be useful?

larsoner:
I don't think we can/should force people to use IPython instead of plain Python. But yes a feature request would be useful, feel free to open one with the PySide6 folks, to me it's a major (deal-breaking) deficiency compared to PyQt6 and I imagine a lot of other people will be in the same boat

hoechenberger:
> @hoechenberger any idea if they plan to build PyQt6?

No plans that I know of. You could ask on the PySide feedstock maybe? But I doubt anyone will be willing to put in the effort tbh...

cbrnr:
It will probably be easier to ask the PySide dev to consider adding an input hook.

hoechenberger:
@cbrnr I kind of lost track. I believe you filed a bug report upstream, right? Any updates on this?

cbrnr:
Yes, here: https://bugreports.qt.io/browse/PYSIDE-2192

It seems like they are working on it? https://codereview.qt-project.org/c/pyside/pyside-setup/+/454254

hoechenberger:
> Yes, here: https://bugreports.qt.io/browse/PYSIDE-2192
> 
> 
> 
> It seems like they are working on it? https://codereview.qt-project.org/c/pyside/pyside-setup/+/454254

Seems it's stalled. Maybe we should ping them...

cbrnr:
> Seems it's stalled. Maybe we should ping them...

Yeah, not sure, in general I don't like "hey, any update on this, when will this be ready" kind of pings.

hoechenberger:
My impression was that the implementation is already there and it's just pending review, so maybe folks just forgot about it? But I see what you mean

cbrnr:
If it's really like that I can ask if there is something I could do to help (e.g. test). I think that's an OK ping.

hoechenberger:
Still waiting for https://codereview.qt-project.org/c/pyside/pyside-setup/+/454254, it's stalled

cbrnr:
I don't remember exactly our discussion, but MNELAB should work fine even without this patch. Installing both PySide6 and PyQt6 in parallel should also work.

hoechenberger:
No, we had trouble installing both in parallel, unfortunately. And Matplotlib doesn't work with PySide6 without the patch adding the input hook

cbrnr:
Do you remember what exactly the problem was? I always have both packages installed and everything works.

hoechenberger:
> I always have both packages installed and everything works.

Yes but you don't work with conda-forge packages, do you?

cbrnr:
No. But I can try if you want.

cbrnr:
@hoechenberger I can install both `pyside6` and `pyqt` in parallel with no errors. The only thing that happens is that `pyside6` get downgraded from `6.5.0` to `6.4.3`, not sure why. It also seems like `pyqt` still installs `PyQt5` (`5.15.7`), there is no version `6`.

cbrnr:
I ran MNELAB and imported both PySide6 and PyQt5.

hoechenberger:
There's a test case here in this branch in build.yml

larsoner:
Any chance to transition to `qtpy` in MNELAB?

cbrnr:
> Any chance to transition to `qtpy` in MNELAB?

I've used `qtpy` previously, but switched away for three reasons:

1. I did not want to support Qt versions < 6 anymore (there are some weird bugs like font rendering that are not going to be fixed).

2. PySide6 has some nice features that PyQt6 doesn't offer, like more Pythonic names.

3. PyQt6 is licensed under the GPL, whereas PySide6 uses the LGPL. I'm not an expert, but the LGPL seems like the much better option here, because it is clear that we can just use it without having to worry about the license of MNELAB. For GPL, although BSD is compatible, I'm not entirely sure that we can use it without having to change our own license to GPL.

4. We still wanted to include Qt bindings as a requirement, so people don't have to figure out why MNELAB does not work after `pip install mnelab`. Just including e.g. `PySide6` defeats the purpose of `qtpy`, and I did not want a brittle custom solution in `setup.py` (which BTW wouldn't even be possible now that we switched to `pyproject.toml`). I did not want users to install with `pip install mnelab[pyside6]`, because I think almost no one knows how this works (I know it would be documented, but few people read the docs and would just use what they know from other packages, which is `pip install mnelab`).

In summary, I'm open to switching back, but these issues need to be addressed. The second one is not a real issue, because AFAIR I'm not even using PySide6-specific features.

hoechenberger:
This should work now:

https://github.com/conda-forge/qt-main-feedstock/pull/167

hoechenberger:
My understanding is that now, thanks to https://github.com/conda-forge/qt-main-feedstock/pull/167, parallel installation of Qt5 and Qt6 should work.

Since we still don't have PyQt6 on conda-forge, we have to rely on PySide6. But the following issue / PR is stalled:
https://codereview.qt-project.org/c/pyside/pyside-setup/+/454254

@cbrnr Feel like bumping this again?

cbrnr:
Yes. Here's the corresponding issue: https://bugreports.qt.io/browse/PYSIDE-2192

cbrnr:
@hoechenberger maybe you (and others) could vote for this issue to show that it is important?

cbrnr:
Thanks! But unfortunately, the answer is not promising. I don't even see the error on Windows. I mean, I probably wouldn't be able to help anyway, but do we know someone who could?

hoechenberger:
I'm really surprised this issue isn't high-prio for them. Are we potentially missing something obvious here? Shouldn't there be literally masses of users running into this problem?

cbrnr:
Hard to say. On macOS, people get to use the `macosx` backend by default, which works great. Not sure about Windows and Linux, but maybe people just use PyQt? The problem is specific to conda-forge and Qt 6, which is only available as PySide6 there. If you use pip, then you also have PyQt6. If you are happy with Qt 5, PyQt5 is also available on conda-forge. So yes, maybe really a niche problem.

hoechenberger:
@cbrnr Any changes here..?

cbrnr:
Not that I know of. My list of caveats still applies, if you mean switching to qtpy.

PR #165, opened 2022-12-14
Use Matplotlib's Qt5Agg backend by default

### LINKED ISSUE #164: Use MPL Qt backend by default on all platforms
Even if we end up not merging #163, we should apply the following commit to `main`:
https://github.com/mne-tools/mne-installers/pull/163/commits/f4f1843f7aeeaa8337eef9219e310a6b82abec03

hoechenberger:
Fixes #164

### DIFF
--- a/assets/post_install_linux.sh
+++ b/assets/post_install_linux.sh
@@ -18,6 +18,9 @@ done
 echo "ℹ️ Configuring Python to ignore user-installed local packages."
 ${PREFIX}/bin/conda env config vars set PYTHONNOUSERSITE=1
 
+echo "ℹ️ Configuring Matplotlib to use the Qt5 backend by default."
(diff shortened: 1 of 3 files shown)


hoechenberger:
@larsoner Feel free to merge if happy & once green

larsoner:
I'm not totally convinced, as this will take precedence over even a .matplotlibrc file if someone has it. If we do it at all, I'd be inclined to only do it on macOS since that's where it's not a Qt backend by default anyway.

cbrnr:
This change only affects our installers, right? I'd still vote for leaving the macOS default backend as is (i.e. `osx`, which is the Matplotlib default for macOS). I have had zero problems with the `osx` backend and also feel that it is faster in some situations (don't have hard numbers though).

PR #173, opened 2023-02-17
Add pyRiemann and update several other packages

hoechenberger:


hoechenberger:
@larsoner I'd love to retain all individual commits here as we merge into `main`, but that would not work without adding a "merge commit", right? Which is not really what I want…

hoechenberger:
I think I'll do individual PRs for each change

hoechenberger:
Superseded by dependency updates in #174, will close and open a new PR just for pyriemann

PR #168, opened 2023-01-26
Include MNE-ICALabel (and PyTorch) in the Windows installers

hoechenberger:

PR #174, opened 2023-02-21
MAINT: Update deps

larsoner:
Separate dep updates and boilerplate `_1` stuff from #168 and #173, and also make sure the installers build with the current stack

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
           - os: macos-11
             arch: arm64
 
-          - os: ubuntu-18.04
+          - os: ubuntu-20.04
             arch: x86_64
 
           - os: windows-2019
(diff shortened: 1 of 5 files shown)


larsoner:
Ubuntu 18.04 keeps encountering an internal error, so I'm going to take this opportunity to try to switch `18.04, 20.40` -> `20.04, 22.04`

larsoner:
@hoechenberger feel free to squash+merge without requirements being met, then we can change the requirements (remove 18.04 test, add 22.04 test)

larsoner:
FYI the `9.2.6` VTK thing we can probably undo once https://github.com/conda-forge/mayavi-feedstock/pull/73 is merged. We could wait to see and push that change here if we want...

hoechenberger:
Oh so VTK was the culprit! 🤦‍♀️ 

Thanks for figuring that out!

hoechenberger:
Thanks @larsoner 

Now let's iterate from here :)

PR #175, opened 2023-02-21
MAINT: VTK

larsoner:
Let's see if it works now


hoechenberger:
Dependency resolution issue for ARM

larsoner:
It's not ARM actually it's x86_64. No idea where the conflict actually is, but `Package openssl conflicts for:` is a "good" candidate. Maybe we just wait to see if this resolves itself at some point...

PR #183, opened 2023-04-03
ENH: Sign Windows binary inplace

larsoner:
We should be able to sign inplace now: https://github.com/SSLcom/actions-codesigner/issues/1

This PR should do it, but I expect some timeout failures we'll need to debug before 1.4. No idea why it took so long on a couple of platforms when I was testing in an isolated branch in this repo, we'll see if it happens again.


larsoner:
I vaguely recall somewhere in the conda or napari docs saying that pinning versions can speed up resolution. Given that resolution took ~30 minutes on macOS and we can get our list of outdated dependencies using `python tests/test_outdated.py` in ~1 min, I'm seeing what happens if I pin everything. Hopefully it runs much faster...

larsoner:
By pinning:

- macOS: 30 min down to 9 min
- macOS ARM: 20 min down to 9 min
- Ubuntu: 24 min down to 5 min
- Windows: 15 min down to 11 min

recipes/mne-python_1.4/construct.yaml
```diff
@@ -0,0 +1,143 @@
...
+welcome_file: ../../assets/welcome.rtf
+readme_text: ""
+conclusion_file: ../../assets/conclusion.rtf
+
+initialize_by_default: False
+register_python_default: False
+
+# default_prefix will be ignored by macOS .pkg installer!
+default_prefix: ${HOME}/mne-python/1.4.0_0                          # [linux]
+default_prefix: "%USERPROFILE%\\mne-python\\1.3.1_0"                # [win]
```
hoechenberger, line 22:
```suggestion
default_prefix: "%USERPROFILE%\\mne-python\\1.4.0_0"                # [win]
```

hoechenberger:
> By pinning:
> 
> * macOS: 30 min down to 9 min
> * macOS ARM: 20 min down to 6 min
> * Ubuntu: 24 min down to 5 min
> * Windows: 15 min down to 11 min

Amazing improvements!

I'm just worried we might run into dependency resolution errors someday if we start pinning too many things. But as long as it works, let's keep it!

larsoner:
I did in 545caad actually but it was fairly easy to resolve. Agreed we should undo it later if needed. FWIW it also gives us some indication of when packages are not all up to date across all architectures due to conflicts, which is otherwise hidden/implicit. In this case currently it's only fsleyes on windows and vtk on Mac non-ARM.

recipes/mne-python_1.4/construct.yaml
```diff
@@ -0,0 +1,145 @@
...
+  # - mne-installer-menus =1.2dev0=*_20221007
+  # - mne-bids =0.11dev0=*_20221007
+  # TODO: ⛔️ ⛔️ ⛔️ DEV BUILDS STOP: CHANGE BEFORE RELEASE! ⛔️ ⛔️ ⛔️
+
+  # TODO: Once 1.4 is actually released, bump these
+  - mne =1.3.1=*_0
+  - mne-installer-menus =1.3.1=*_0
+  - mne-bids =0.12.0
+  - mne-bids-pipeline =1.2.0
+  - mne-qt-browser =0.4.0
```
larsoner, line 70:
```suggestion
  - mne-qt-browser =0.5.0
```

PR #185, opened 2023-04-26
ENH: Add git and make for Windows

larsoner:
I missed these when working on Windows recently. Also update outdated dependencies.

### DIFF
--- a/recipes/mne-python_1.4/construct.yaml
+++ b/recipes/mne-python_1.4/construct.yaml
@@ -80,13 +80,13 @@ specs:
   - autoreject =0.4.1
   - pyriemann =0.4
   - pyprep =0.4.2
-  - openmeeg =2.5.5
+  - openmeeg =2.5.6
   # Python
   - python =3.10.8
-  - pip =23.0.1
+  - pip =23.1.2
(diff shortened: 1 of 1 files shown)


recipes/mne-python_1.4/construct.yaml
```diff
@@ -134,6 +134,8 @@ specs:
   - plotly =5.14.1
   - vtk =9.2.6  # [arm64 or not osx]
   - vtk =9.2.5  # [osx and not arm64]
+  - git =2.40.1  # [win]
+  - make =4.3  # [win]
```
larsoner, line 138: These are just `[win]` because I think macOS and Windows people can be expected to have them on their own somehow (`brew` or `apt`). Windows people are unlikely unless they're using msysgit which is not super likely, and even if they are it probably/hopefully won't cause problems. (I installed these and was able to use them.)

hoechenberger: Yes this makes sense. On macOS you're likely to have the Xcode CLI tools installed, which ship git

hoechenberger:
Thanks Eric

PR #186, opened 2023-05-03
ENH: Bump versions

larsoner:
- Tries 3.11
- Adds stub for Numba where we should add it once it lands
- Needs latest dev version https://github.com/conda-forge/mne-feedstock/pull/118 to land and hit CDNs (this one does not require numba)
- By that time, `mne-gui-addons` should also be in CDNs, which I've added here
- I'm trying to remove some pins for `fsleyes` and `vtk` to see if they're magically resolved

First run will undoubtedly fail, I'll restart in an hour or two

### DIFF
--- a/recipes/mne-python_1.4/construct.yaml
+++ b/recipes/mne-python_1.4/construct.yaml
@@ -77,7 +77,7 @@ specs:
   - mne-ari =0.1.2
   - mne-kit-gui =1.1.0
   - mne-icalabel =0.4  # [not win]
-  - mne-gui-addons =0.2.0
+  - mne-gui-addons =0.1.0
   - autoreject =0.4.1
   - pyriemann =0.4
   - pyprep =0.4.2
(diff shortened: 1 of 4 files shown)


hoechenberger (approved):

hoechenberger:
Thanks Eric! Shall this be merged before the release of MNE 1.4 or do you want to keep it open until then?

larsoner:
Marking as draft because there's a bunch of stuff I'll need to comment out and wait on...

larsoner:
Looks like our 3.11 blockers are:

- [ ] numba
- [ ] tensorpac (needs numba)
- [ ] emd (needs new release without numpy pin)
- [ ] sleepecg (needs numba, maybe tensorflow?)
- [ ] yasa (needs sleepecg and tensorpac)

I hope these clear up by Friday but if not, I'm tempted to release 1.4 using Python 3.10. It might be safer anyway given there are these many packages with issues, there are probably others that users will want to install into their envs and not be able to :(

hoechenberger:
@larsoner We can make another release in a few weeks with 3.11, no need to force anything now

hoechenberger (approved):

PR #188, opened 2023-05-17
BUG: Add tests for setuptools_scm versions, fix mne-qt-browser

larsoner:
To help check https://github.com/mne-tools/mne-qt-browser/issues/165

Eventually we could probably assert actual version numbers based on our pinning file and such. But for now this is good enough I think

Should fail until https://github.com/conda-forge/mne-qt-browser-feedstock/pull/24 is in and CDNs update

### DIFF
--- a/recipes/mne-python_1.4/construct.yaml
+++ b/recipes/mne-python_1.4/construct.yaml
@@ -88,8 +88,8 @@ specs:
   - mamba =1.4.2
   - openblas =0.3.23
   - jupyter =1.0.0
-  - jupyterlab =3.6.3
-  - ipykernel =6.23.0
+  - jupyterlab =4.0.0
+  - ipykernel =6.23.1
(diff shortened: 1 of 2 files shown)


hoechenberger:
I don't understand why we'd need this – shouldn't dependency resolution fail if we cannot get the package version that we requested in the recipe? Or has this to do with `setuptools-scm`?

larsoner:
> I'm okay with having this here as a temporary check until we start testing for the actual version numbers we put in the recipe, like you suggested

Might as well be permanent -- conda's package version numbers are not guaranteed to match the actual `module.__version__`, so we might as well check the `setuptools_scm` packages that we have to make sure that they do.

larsoner:
Will merge and cut a 1.4.0_1 release since mne-qt-browser Epochs browsing is broken on 1.4.0_0

PR #193, opened 2023-07-06
Add pyxdf

### LINKED ISSUE #192: Include pyxdf
I think it would be useful to include `pyxdf` for importing XDF files.

cbrnr:
Fixes #192.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,7 +112,7 @@ jobs:
 
       - name: Hotfix for NSIS
         run: |
-          mamba install nsis=3.01 conda=4.12a pydantic<2.0a0
+          mamba install nsis=3.01 conda=4.12a "pydantic<2.0a0"
         if: ${{ runner.os == 'Windows' }}
 
       - name: Extract version information
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,7 +112,7 @@ jobs:
 
       - name: Hotfix for NSIS
         run: |
-          mamba install nsis=3.01 conda=4.12a
+          mamba install nsis=3.01 conda=4.12a pydantic<2.0a0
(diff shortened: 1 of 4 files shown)


larsoner:
You will need to upgrade the out of date ones then cross your fingers that the resolvers are happy. If they're not we have to isolate the problem and add an out-of-date exception. It's a bit tedious but it makes sure we're as up to date as possible

larsoner:
Looking at

- https://github.com/mne-tools/mne-installers/blob/main/environment.yml
- https://anaconda.org/napari/menuinst/files?page=2
- https://anaconda.org/conda-forge/nsis/files

They released a new version of `nsis` 2 months ago, maybe we need to pin to build _1 instead of allowing _2. Another possibility I thought might be `menuinst` but we already pin that one.

cbrnr:
The "NSIS Hotfix" job consists of this line (which will only run on Windows):

```
mamba install nsis=3.01 conda=4.12a
```

We're already pinning `nsis=3.01`. Do you mean that we should try using the latest version? I don't know what that hotfix does, why is it pinning these two packages in the first place?

jaimergp:
I think it must be the new pydantic. Pin it to `pydantic <2.0a0`.

cbrnr:
Apparently `mamba install pydantic<2.0.0a1` doesn't work (sorry, I don't use `mamba`/`conda`)...

larsoner:
I *think* that might be a standard BASH issue with `<`, I'll push quotes to see if it helps

cbrnr:
> I think that might be a standard BASH issue with <, I'll push quotes to see if it helps

Ah, of course! Good catch! It doesn't solve the original problem though 😢.

hoechenberger:
Oh no, what a pain :(

hoechenberger (approved):

cbrnr:
Thanks @larsoner!

PR #196, opened 2023-07-22
ENH: Use 3.11

larsoner:
Now that https://github.com/conda-forge/sleepecg-feedstock/pull/22 is in let's see if it works!

Also sets us up for 1.5 because we'll probably release that soon


larsoner:
Worked for everything but Linux! On Monday I'll look to see where the conflict is coming from

hoechenberger:
This means we can start shipping TensorFlow now too

larsoner:
Held up by https://github.com/conda-forge/vtk-feedstock/pull/302 (and maybe https://github.com/conda-forge/mayavi-feedstock/pull/75)

larsoner:
Now held up by https://github.com/conda-forge/xorg-libxt-feedstock/issues/14

larsoner:
All green, in it goes!!!

hoechenberger (commented):
very nice

PR #195, opened 2023-07-13
MAINT: Use bundle_tools_3 and fix various bugs

larsoner:
A naive attempt at https://github.com/conda/constructor/pull/474#issuecomment-1634249245

- [x] Get CIs green
- [x] Put anything back that macOS needs
- [x] Check installer on Windows
  - [x] Single-user
  - [x] All-user
- [x] Check installer on Linux
- [x] Add developer dependencies
- [x] Wait for https://github.com/conda-forge/blas-feedstock/pull/104 and rerun
- [x] Wait for https://github.com/conda-forge/mne-features-feedstock/pull/4 and rerun
- [ ] Check installer on macOS
  - [x] larsoner (M1)
    - [x] user install
    - [x] system-wide install
  - [ ] hoechenberger
  - [ ] cbrnr

Closes #204
Closes #191
Closes #190
Closes #182
Closes #181
Closes #131

PR #205, opened 2023-09-06
MAINT: Bump to 1.5.1

larsoner:


larsoner:
Putting in a fix for https://github.com/mne-tools/mne-installers/actions/runs/6100615960/job/16555293838 where `MNE_INSTALL_PREFIX` was not added properly to the env

.github/workflows/build.yml
```diff
@@ -153,15 +156,21 @@ jobs:
...
 
       - name: Check installer signature (macOS)
-        if: ${{ runner.os == 'macOS' && github.event_name != 'pull_request' }}
+        if: ${{ runner.os == 'macOS' }}
         run: |
           # Installer package
           set -eo pipefail
-          pkgutil --check-signature ${MNE_INSTALLER_NAME}
+          if [[ "${{ github.event_name }}" != "pull_request" ]]; then
+            pkgutil --check-signature ${MNE_INSTALLER_NAME} || exit 1
```
larsoner, line 164: FYI @hoechenberger this is the only step that *fails* when in PR mode, so I removed the conditinal in the GH Actions step `if` above and put it here instead. The other adjustments below are just path failures from #195

PR #210, opened 2023-09-27
ENH: Add profiling tools

larsoner:
1. Add profiling tools for developer use.
2. Update deps.
3. Remove windows signing. It was too expensive and it's easy enough on Windows to say "install anyway" (much easier than on macOS).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -195,7 +195,6 @@ jobs:
           # Staple the notarization certificate onto it
           xcrun stapler staple ${MNE_INSTALLER_NAME}
 
-      # Note: if this changes, it needs to be changed in the Windows signing step as well
       - name: Calculate SHA256 hash of installer package
         run: |
           shopt -s nullglob  # Fail if the following pattern yields no results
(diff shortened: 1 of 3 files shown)


larsoner:
I'm going to mark this for merge-when-green since I think this is uncontroversial assuming it works, but @hoechenberger feel free to look and I'm happy to adjust in a follow-up PR if you find tweaks.

hoechenberger:
> 3\. and it's easy enough on Windows to say "install anyway"

I'd rather we didn't train users to simply disregard security warnings when running stuff they randomly downloaded off the internet

Is there no way to obtain a more affordable signing cert?

larsoner:
> Is there no way to obtain a more affordable signing cert?

Not that I know of and I searched a bit originally. It was > $1k per year and the MNE-Python R01 got some budget cuts with the move to Aalto :(

larsoner:
> I'm paying for the macOS cert not because using an unsigned app is more inconvenient, but because I believe it's good practice to sign stuff you ship

Totally agree. I'm just saying I think the damage / additional thoughts/actions on the part of users is less bad for Windows than it is for macOS. Agreed we should have signed Windows, and we would if it were as affordable as it is for macOS

drammock:
Have either of you looked into https://www.sigstore.dev/ (from Linux Foundation)?

hoechenberger:
> It was > $1k per year

I didn't know it was so ridiculously expensive! This is 10 times as much as the Apple cert… ouch

larsoner:
@drammock no I haven't. From a very brief and naive look I'm not sure it solves the problem of OS-based hurdles / warnings during app installation for macOS or Windows, though (what we're dealing with here). In a related issue I saw something about SignPath, which we should look into:

- https://github.com/sigstore/fulcio/issues/250#issuecomment-1697130810
- https://signpath.org/
- https://about.signpath.io

Not sure if we'd qualify to get it for free but maybe worth looking into...

PR #211, opened 2023-09-27
MAINT: Reorganize to single folder structure

### LINKED ISSUE #209: MAINT: Migrate to standard version control procedures
It's a bit weird to me to have `recipes/mne_python_*.*/` rather than a single `recipes/mne_python/`.

- Most old recipes very well might not work anymore.
- I don't think we'll ever really support multiple versions simultaneously, but even if we do it seems more standard to do this with branches (like `maint/1.5` and `maint/1.6`) than it does with folders.
- Having multiple versions ruins our `git blame` for lines (for example [our latest construct.yaml](https://github.com/mne-tools/mne-installers/blame/main/recipes/mne-python_1.5/construct.yaml) only has changes going back ~3mo even though some lines are effectively much older)

@hoechenberger if you agree with migrating to a more standard version control scheme under a single `recipes/mne_python/`, we could try to migrate gracefully by merging backward (rename 0.23 to root `mne_python`, then merge 0.24 into `mne_python`, then 0.25 into 0.23, etc. in single PRs). That way we at least get to see what was changed in each version release. (I'll get all `git blame` in terms of authorship but at least it'll be clearer when stuff was changed.)

larsoner:
**THIS PR SHOULD BE MERGED BY REBASE NOT SQUASH-MERGED**

1. Move `mne-python_0.23` to `mne-python`
2. Move 0.24, 1.0, ..., 1.5's `construct.yaml` to `recipes/mne-python/construct.yaml` in individual commits.
3. Update code to accommodate new filename structure

We should merge #210 first then merge this one by rebase

Closes #209


larsoner:
I'll keep force-pushing this until green, but maybe we'll be there with the latest commit 🤞 Marking for rebase-and-merge when green to make sure the history is reasonable!

larsoner:
Blame actually looks okay! https://github.com/mne-tools/mne-installers/blame/main/recipes/mne-python/construct.yaml

hoechenberger:
Amazing job

PR #213, opened 2023-09-30
MAINT: Move to Python 3.12

larsoner:
An auto rebuild triggered for OpenMEEG in https://github.com/conda-forge/openmeeg-feedstock/pull/46 and I figured we might as well get this on our radar. Locally it still needs matplotlib so doesn't get very far:
```
$ ./tools/build_local.sh 
...
Channels:
 - conda-forge/label/python_rc
 - conda-forge
Platform: osx-arm64
Collecting package metadata (repodata.json): done
Solving environment: failed
...
Could not solve for environment specs
The following packages are incompatible
├─ python 3.12.0rc3**  is installable and it requires
│  └─ python_abi 3.12.* *_cp312, which can be installed;
└─ seaborn 0.12.2**  is not installable because it requires
   └─ seaborn-base >=0.12.2,<0.12.3.0a0 , which requires
      └─ matplotlib-base >=3.1,!=3.6.1  but there are no viable options
         ├─ matplotlib-base [3.3.2|3.3.3|...|3.7.3] would require
         │  └─ python_abi 3.8.* *_cp38, which conflicts with any installable versions previously reported;
         ├─ matplotlib-base [3.3.2|3.3.3|...|3.8.0] would require
         │  └─ python_abi 3.9.* *_cp39, which conflicts with any installable versions previously reported;
         ├─ matplotlib-base [3.4.3|3.5.0|...|3.8.0] would require
         │  └─ python_abi 3.10.* *_cp310, which conflicts with any installable versions previously reported;
         └─ matplotlib-base [3.6.2|3.6.3|...|3.8.0] would require
            └─ python_abi 3.11.* *_cp311, which conflicts with any installable versions previously reported.
```

### DIFF
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -60,7 +60,7 @@ channels:
 
 specs:
   # Python
-  - python =3.12.1
+  - python =3.12.2
   - pip =24.0
   - conda =23.10.0
   - mamba =1.5.3
@@ -102,8 +102,6 @@ specs:
   - jupyter =1.0.0
   - jupyterlab =4.1.2
   - ipykernel =6.29.2
-  # https://github.com/Anaconda-Platform/nb_conda_kernels/issues/230
-  # - nb_conda_kernels =2.3.1
   - spyder-kernels =2.5.0
   - spyder =5.5.1
   - darkdetect =0.8.0
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -86,6 +86,7 @@ specs:
   - mne-features =0.3
   - mne-rsa =0.9
   - mne-ari =0.1.2
+  # https://github.com/conda-forge/mayavi-feedstock/pull/77
   # - mne-kit-gui =1.2.0
   - mne-icalabel =0.6.0
   - mne-gui-addons =0.1.0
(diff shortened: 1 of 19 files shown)


larsoner:
Still waiting on Numba and dipy (at least) here

recipes/mne-python/construct.yaml
```diff
@@ -96,12 +97,13 @@ specs:
...
   # BLAS
-  - openblas =0.3.25
+  - openblas =0.3.26
   - libblas =3.9.0=*openblas
   - jupyter =1.0.0
   - jupyterlab =4.1.2
   - ipykernel =6.29.2
-  - nb_conda_kernels =2.3.1
+  # https://github.com/Anaconda-Platform/nb_conda_kernels/issues/230
+  # - nb_conda_kernels =2.3.1
```
hoechenberger, line 106: Actually I don't think we need this anyway, it's only required to select kernels from a different conda environment. I believe we should be fine without this in the installer

hoechenberger:
@larsoner I see you removed TensorFlow, was this necessary to get the package to build? TF is an optional, yet important dependency of SleepECG. I believe classification without TF is either impossible or extremely slow. @cbrnr at least told me  something along those lines, if I recall correctly

tests/test_outdated.py
```diff
@@ -24,16 +24,13 @@ class Package:  # noqa: D101
 
 
 allowed_outdated: set[str] = {
-    "python",  # ignore 3.12
     "conda",
     "mamba",  # conda/mamba conflict with tensorflow and VTK
```
hoechenberger, line 28: Do you think we could resolve this somehow?

Also I started transitioning away from `mamba` again in my personal dev environments, since `conda` now uses the `libmamba` solver by default

… this just to suggest that we may consider simply ditching `mamba` if it causes trouble!

cbrnr:
Yes, SleepECG needs TF for classification.

larsoner:
TF is not available on Python 3.12. So we'll need to revert this if it's essential

PR #215, opened 2023-10-16
Add mne-lsl

mscheltienne:
Adding `mne-lsl` to the installers.
Should we keep `mne-realtime`, people mentioned it was not working with the latest version of MNE anyway?

### DIFF
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -61,10 +61,10 @@ channels:
 specs:
   # Python
   - python =3.11.5
-  - pip =23.2.1
-  - conda =23.7.4
+  - pip =23.3
+  - conda =23.9.0
   - mamba =1.5.1
-  - conda-libmamba-solver =23.7.0
+  - conda-libmamba-solver =23.9.1
   # MNE ecosystem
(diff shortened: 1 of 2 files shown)


hoechenberger:
What's the difference between realtime and LSL? Can MNE-LSL replace MNE-realtime for LSL users?

drammock:
> What's the difference between realtime and LSL? 

MNE-LSL is a rewrite of much of MNE-Realtime functionality, but built on top of LSL from the start (MNE-Realtime was built around FieldTripBuffer, with limited LSL support added later). 

> Can MNE-LSL replace MNE-realtime for LSL users?

yes, that is the intent.

mscheltienne:
It's still a WIP on which I work on when I have the time, but it is functional and already simplifies a lot the interaction with an LSL stream.

hoechenberger:
Thanks @drammock and @mscheltienne!

I just saw that you were planning to deprecate the MNE-Realtime package anyway, so from my end I have no objections to removing it and adding MNE-LSL to the installer.

hoechenberger (changes requested):
Please add an import test to `test_imports.py` and bump the versions of all other outdated packages in a separate commit. I think then it'll be ready to merge.

larsoner:
`mne-realtime` should in principle continue to work. Let's keep it for a release or two

hoechenberger (approved):

PR #178, opened 2023-02-22
Include MNE-ICALabel and PyTorch in Windows installers

hoechenberger:
New attempt after the failed #168

PR #219, opened 2023-10-30
Bump mne-icalabel

### LINKED ISSUE #218: Package MNE-ICALabel for Windows, too
Since 0.5.1, MNE-ICALabel can run via ONNX and doesn't require PyTorch anymore. This should allow us to easily deploy it on Windows, too

cc @mscheltienne

mscheltienne:
Version 0.5.1 depends on `onnxruntime` instead of `pytorch`.

Closes #178
Closes #218

### DIFF
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -64,7 +64,7 @@ specs:
   - pip =23.3.1
   - conda =23.9.0
   - mamba =1.5.2
-  - conda-libmamba-solver =23.9.2
+  - conda-libmamba-solver =23.9.3
   # MNE ecosystem
   # TODO: ⛔️ ⛔️ ⛔️ DEV BUILDS START: CHANGE BEFORE RELEASE! ⛔️ ⛔️ ⛔️
   # - mne =1.4dev0=*_20230503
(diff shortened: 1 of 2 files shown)


recipes/mne-python/construct.yaml
```diff
@@ -86,25 +86,25 @@ specs:
...
   - jupyter =1.0.0
   - jupyterlab =4.0.7
-  - ipykernel =6.25.2
+  - ipykernel =6.26.0
   - nb_conda_kernels =2.3.1
   - spyder-kernels =2.4.4
   - spyder =5.4.5
   - darkdetect =0.8.0
-  - qdarkstyle =3.1
+  - qdarkstyle =3.2
```
hoechenberger, line 107: This seems to be incompatible with Spyder

mscheltienne: Thanks, I was a bit discourage by the length of the traceback:

```
Could not solve for environment specs
The following packages are incompatible
├─ fsleyes 1.9.0**  is installable with the potential options
│  ├─ fsleyes 1.9.0 would require
│  │  └─ __linux, which is missing on the system;
│  ├─ fsleyes 1.9.0 would require
....
```

and missed the relevant information above.

hoechenberger:
Thanks @mscheltienne

PR #225, opened 2023-11-14
MAINT: Check against mne pip option

larsoner:
Related to https://github.com/mne-tools/mne-python/issues/11831. We should check that `mne[test,test_extra,full,doc]` is installed by our installer.

Will fail until https://github.com/mne-tools/mne-python/pull/12209 is merged, should say something like:
```
AssertionError: Unexpected unmet dependencies:
Sphinx-6.2.1
mne-bids-0.14.dev0
mne-connectivity-0.6.0.dev0
mne-gui-addons-0.2.0.dev9+gc245654
pydata-sphinx-theme-0.13.3
sphinx-gallery-0.15.dev0
```

### DIFF
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -94,6 +94,7 @@ specs:
   - pyriemann =0.5
   - pyprep =0.4.3
   - openmeeg =2.5.6
+  - python-neo =0.12.0
   # BLAS
   - openblas =0.3.24
   - libblas =3.9.0=*openblas
@@ -149,6 +150,7 @@ specs:
   - make =4.3  # [win]
   - ipywidgets =8.1.1
   - pyvista =0.42.3
+  - pyvistaqt =0.11.0
   - trame =3.3.0
   - trame-vtk =2.6.1
   - trame-vuetify =2.3.1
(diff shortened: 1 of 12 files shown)


recipes/mne-python/construct.yaml
```diff
@@ -112,9 +113,8 @@ specs:
   - xlrd =2.0.1
   # Statistics
   - pingouin =0.5.3
-  - pycircstat =0.0.2
```
larsoner, line 115: This appears to be unmaintained, relies on `nose` for example and meaningful commits are from a while ago

https://github.com/circstat/pycircstat

It creates a conflict, so let's remove it https://github.com/mne-tools/mne-installers/actions/runs/6871571108/job/18688630912?pr=225

hoechenberger: Ok!

larsoner:
The GHA build/test architecture had a couple big drawbacks:

1. The file was getting very long
2. All builds needed to complete before any tests were run

I thus refactored the action by:

1. Splitting `build, test` into `build_macos, build_linux, build_windows, test_macos, test_linux, test_windows` jobs. 
2. Moving shell commands in building and testing to new `tools/*.sh` scripts
3. Removing the `name:` for any steps where it was obvious from the script name and/or action what was happening

This ends up with some redundancy, but the flow is much more readable (way fewer `if` clauses), and it ends up at about half the number of lines compared to before. And now if one build finishes first -- like the Linux one -- its tests can start running immediately for faster feedback.

hoechenberger:
@larsoner please go ahead if you're happy, I won't have time to review today

larsoner:
Linux *tests* completed before the macOS and Windows *builds* even completed :heart_eyes:

PR #227, opened 2023-11-20
MAINT: Updates to dependencies

larsoner:
Draft until we can actually change to 1.6, just making sure stuff is green in the meantime

### DIFF
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -174,7 +174,7 @@ specs:
   - twine =4.0.2
   # Doc building
   - numpydoc =1.5.0
-  - pydata-sphinx-theme =0.14.3
+  - pydata-sphinx-theme =0.13.3
   - graphviz =8.1.0
   - python-graphviz =0.20.1
   - selenium =4.15.2
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -175,7 +175,7 @@ specs:
   # Doc building
   - numpydoc =1.5.0
   - pydata-sphinx-theme =0.14.3
-  - graphviz =9.0.0
+  - graphviz =8.1.0
   - python-graphviz =0.20.1
   - selenium =4.15.2
   - sphinx-design =0.5.0
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -64,7 +64,7 @@ specs:
   - pip =23.3.1
   - conda =23.10.0
   - mamba =1.5.3
-  - conda-libmamba-solver =23.11.0
+  - conda-libmamba-solver =23.11.1
   # MNE ecosystem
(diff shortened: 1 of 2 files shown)


recipes/mne-python/construct.yaml
```diff
@@ -75,7 +75,7 @@ specs:
   - mne-installer-menus =1.5.1=*_0
   # For testing purposes with build_local.sh, you can comment out all deps
   # below for speed, and change mne to mne-base above
-  - mne-bids =0.13.0
+  - mne-bids =0.14
```
larsoner, line 78: cc @sappelhoff it seems like something went a bit off with the latest mne-bids release on conda-forge as it's 0.14 not 0.14.0. It's no problem here but figured you might want to look into this as it seems non-standard

sappelhoff: Thanks for the ping, but I don't think I understand the problem 🤔 AFAIK we always had versions like this:

- https://pypi.org/project/mne-bids/#history
- https://github.com/conda-forge/mne-bids-feedstock/commits/main/recipe/meta.yaml

I.e., we drop the trailing patch version of `.0` if it is a zero, and only append if it is >= `.1`

larsoner: Seems like it might be non-standard behavior to do that. Or at least I noticed that MNE-BIDS is the only package in this list that versions that way. Based on a tiny bit of SemVer reading just now it sounds like it's better in general to have the `.0` at the end if it's not a pain to add it

sappelhoff: That's a good point, thanks. I believe we took the versioning system from mne-python though --> isn't that how releases were usually done? The history seems to agree for mne versions <0.17.

It seems like we've missed a "change" in how versions are being done at mne, because since about 0.17 (it's inconsistent on pypi and github), there is always the `.0` patch addendum.

cc @hoechenberger @jasmainak @agramfort 

- https://github.com/mne-tools/mne-python/tags?after=v0.19.1
- https://pypi.org/project/mne/#history

I am not sure what to do now. Shall we just adjust the mne-bids versioning protocol for future versions? So far we haven't had any issues with the current scheme.

hoechenberger: I don't think we ever did a patch release for MNE-BIDS, or did we? I'm okay with leaving things as they are, i.e., just MAJOR.MINOR

In fact, we could even drop the 0. and just use current MINOR as only (MAJOR) version number, would be closer to SemVer anyway :)

hoechenberger: Ah, we had MNE-BIDS 0.11.1 once :)

sappelhoff: > I don't think we ever did a patch release for MNE-BIDS, or did we

we do have two patch releases, one of them being: https://github.com/mne-tools/mne-bids/releases/tag/v0.11.1

I am +1 to adding the `.0` from now on, and being more semver-ish

hoechenberger: sounds good to me!!

PR #231, opened 2023-11-21
Bump pycrostates and mne-lsl

mscheltienne:
Already available on `conda-forge`, surprisingly quick.


mscheltienne:
... or maybe that was too fast :see_no_evil:

larsoner:
Should be no problem to get this in @mscheltienne

PR #233, opened 2023-11-21
Add more logging output to Apple certificate workflow and ensure OpenSSL can handle our Developer certificates

### LINKED ISSUE #232: BUG: macOS certificates broken
@hoechenberger in case you didn't see the top comment from https://github.com/mne-tools/mne-installers/pull/230#issue-2004552478 , the macOS certs have an issue that will make it so we can't release:

https://github.com/mne-tools/mne-installers/actions/runs/6946567994/job/18898366624

hoechenberger:
Closes #232

cc @larsoner 

I would almost bet it's the `openssl` commands that are the issue; they're optional and we could just drop them for now to push out a release quickly. But let's wait and see

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,6 @@ on:  # yamllint disable-line rule:truthy
   push:
     branches:
       - main
-      - macos-certs
     tags:
       - '*'
   pull_request:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,7 @@ on:  # yamllint disable-line rule:truthy
   push:
     branches:
       - main
+      - macos-certs
     tags:
       - '*'
   pull_request:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
           environment-file: environment.yml
       - run: ./tools/extract_version.sh
       - run: ./tools/macos_install_certificates.sh
-        # if: ${{ github.event_name != 'pull_request' }}
+        if: ${{ github.event_name != 'pull_request' }}
         env:
           APPLICATION_CERT_BASE64: ${{ secrets.APPLE_APPLICATION_CERT_BASE64 }}
           APPLICATION_CERT_PASSWORD: ${{ secrets.APPLE_APPLICATION_CERT_PASSWORD }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
           environment-file: environment.yml
       - run: ./tools/extract_version.sh
       - run: ./tools/macos_install_certificates.sh
-        if: ${{ github.event_name != 'pull_request' }}
+        # if: ${{ github.event_name != 'pull_request' }}
         env:
           APPLICATION_CERT_BASE64: ${{ secrets.APPLE_APPLICATION_CERT_BASE64 }}
           APPLICATION_CERT_PASSWORD: ${{ secrets.APPLE_APPLICATION_CERT_PASSWORD }}
--- a/tools/macos_install_certificates.sh
+++ b/tools/macos_install_certificates.sh
@@ -8,7 +8,7 @@ exec 2>&1  # redirect stderr to stdout, so error messages show up in GH Actions 
 APPLICATION_CERT_PATH=$RUNNER_TEMP/application_cert.p12
 INSTALLER_CERT_PATH=$RUNNER_TEMP/installer_cert.p12
 KEYCHAIN_PATH=$RUNNER_TEMP/app-signing.keychain-db
-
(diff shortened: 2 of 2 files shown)


tools/macos_install_certificates.sh
```diff
@@ -13,42 +13,60 @@ echo -n "$APPLICATION_CERT_BASE64" | base64 --decode --output $APPLICATION_CERT_
 echo -n "$INSTALLER_CERT_BASE64" | base64 --decode --output $INSTALLER_CERT_PATH
 
 # create temporary keychain
+echo "🏃 Creating temporary keychains …"
```
hoechenberger, line 16:
```suggestion
echo "🏃 Creating temporary keychain …"
```

tools/macos_install_certificates.sh
```diff
@@ -13,42 +13,60 @@ echo -n "$APPLICATION_CERT_BASE64" | base64 --decode --output $APPLICATION_CERT_
 echo -n "$INSTALLER_CERT_BASE64" | base64 --decode --output $INSTALLER_CERT_PATH
 
 # create temporary keychain
+echo "🏃 Creating temporary keychain …"
 security create-keychain -p "$KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
 security set-keychain-settings -lut 21600 $KEYCHAIN_PATH
 security unlock-keychain -p "$KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
+echo "✅ Done creating temporary keychains."
```
hoechenberger, line 20:
```suggestion
echo "✅ Done creating temporary keychain."
```

hoechenberger:
Yep, that's it. CI run stops with the line:


```
🏃 Running sanity check on our Developer certificates before importing …
```
The following two lines are OpenSSL commands.

Must be something to do with OpenSSL 3, I suppose.

macOS Sonoma ships the new version.

I can reproduce locally:

```shell
❯ openssl pkcs12 -nokeys -passin pass:"" -in developer_id_application.p12 |grep friendlyName
Error outputting keys and certificates
C01E3EDD01000000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:341:Global default library context, Algorithm (RC2-40-CBC : 0), Properties ()
```

```shell
❯ openssl version

OpenSSL 3.1.1 30 May 2023 (Library: OpenSSL 3.1.1 30 May 2023)
```

@larsoner I'll be gone for a couple hours or until tomorrow, would you like to take it from here?

hoechenberger:
@larsoner I just found this:
https://developer.apple.com/forums/thread/734228

I may need to re-export the keys, as I suspected

hoechenberger:
@larsoner This should pass now

before merging, we must revert that one change to the GH actions config

Can you take over please?

hoechenberger:
@larsoner I don't understand this, I'm not seeing any error messages anymore :(

hoechenberger:
I have freshly exported the certificates and added them as GH Actions Secrets

.github/workflows/build.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           environment-file: environment.yml
       - run: ./tools/extract_version.sh
       - run: ./tools/macos_install_certificates.sh
-        if: ${{ github.event_name != 'pull_request' }}
+        # if: ${{ github.event_name != 'pull_request' }}
```
larsoner, line 47: I don't think secrets get passed to PRs so commenting this out isn't enough. You'd need to be pushing to a branch on this repo for example rather than your fork

larsoner:
... testing in https://github.com/mne-tools/mne-installers/tree/macos-certs and https://github.com/mne-tools/mne-installers/actions/runs/6949401613/job/18907440469, sorry about pushing here accidentally

.github/workflows/build.yml
hoechenberger: oooohhhh

larsoner:
Sorry I mean https://github.com/mne-tools/mne-installers/actions/runs/6949434817/job/18907538690

hoechenberger:
@larsoner Ah darn! what a waste of time :) Thanks for the pointer!

Bu I'm off to bed now … :)

larsoner:
https://github.com/mne-tools/mne-installers/actions/runs/6949434817/job/18907538690 got past the certificate stage so I'll just revert this commented-out part here and merge! Sorry I wasn't faster to give you that info

hoechenberger:
@larsoner the new branch you created seems to be working!

Before merging, could you maybe check if you could replace `openssl` with `/usr/bin/openssl`? `openssl` is from `conda-forge`, while the other one is the native binary shipping with macOS. It shouldn't need (and probably doesn't support) the `-legacy` switch

larsoner:
`/usr/bin/openssl` seems to have worked in https://github.com/mne-tools/mne-installers/actions/runs/6949524101/job/18907838481 so I'll force-push here and mark for merge when green

PR #237, opened 2023-12-18
MAINT: Update deps

larsoner:
https://github.com/mne-tools/mne-installers/pull/236 was red, let's see if we can keep things green. FYI @hoechenberger I usually try to push these updates as part of those PRs. It's a good excuse to make sure everything still works rather than having to wait until later. Always feels good to merge things that are green, too.


hoechenberger:
Ah all right, got you, @larsoner! I will consider that in the future

larsoner:
Superseded by other PRs

PR #249, opened 2024-02-18
Start using Apple Silicon, and add test on macOS-13 (Intel)

hoechenberger:
@larsoner I know you love this stuff ;)
I just wanted to get us started

We need to change `extract_version.sh`, I think you have a much better understanding of what needs to be done than I do. Please do feel free to take over if you feel like it!!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,16 +19,11 @@ on:  # yamllint disable-line rule:truthy
 jobs:
   # Build installers
   build_macos:
-    name: Build (macOS ${{ matrix.arch }})
+    name: Build (macOS)
     strategy:
       fail-fast: false
-      matrix:
-        include:
-          - arch: x86_64
-          - arch: arm64
     runs-on: macos-11
     env:
-      MNE_CROSSCOMPILE_ARCH: ${{ matrix.arch }}
       GITHUB_EVENT_NAME: ${{ github.event_name }}
     defaults:
       run:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -126,17 +121,12 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-11, macos-12]
-        arch: [x86_64]
-        # We currently can't/don't test anything about the arm64 build
-        # include:
-        #   - os: macos-11
-        #     arch: arm64
+        os: [macos-11, macos-12, macos-13, macos-14]
```
hoechenberger, line 129: also added macOS-13 here (runs on Intel; macos-14 always runs on M1)

tools/extract_version.sh
```diff
@@ -84,10 +84,6 @@ if [[ "$GITHUB_ACTIONS" == "true" ]]; then
     echo "MNE_INSTALL_PREFIX=${MNE_INSTALL_PREFIX}" >> $GITHUB_ENV
     echo "RECIPE_DIR=${RECIPE_DIR}" >> $GITHUB_ENV
     echo "CONDA_SOLVER=libmamba" >> $GITHUB_ENV
-    if [[ "$MNE_CROSSCOMPILE_ARCH" == "arm64" ]]; then
-        echo "PLATFORM_ARG=--platform=osx-arm64" >> $GITHUB_ENV
-        echo "EXE_ARG=--conda-exe=${CONDA_PREFIX}/standalone_conda/conda.exe" >> $GITHUB_ENV
-    fi
```
hoechenberger, line 90: we shouldn't need this anymore, but I'm not sure about the implications

hoechenberger:
@larsoner I think this is good to go

larsoner:
Awesome @hoechenberger !

PR #269, opened 2024-06-14
MAINT: Update to v1.7.1

larsoner:
Empty commit to start until https://github.com/mne-tools/mne-python/pull/12666 lands, PyPI is pushed, conda-forge is updated. Then I can update to v1.7.1 here and mark for merge-when-green. In the meantime the `autofix.ci` can push an update for outdated deps and we can make sure that's green at least.

### DIFF
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -144,7 +144,7 @@ specs:
   - ipympl =0.9.4
   - seaborn =0.13.2
   - plotly =5.22.0
-  - vtk =9.3.0
+  - vtk =9.2.6
   - git =2.45.2  # [win]
   - make =4.3  # [win]
   - ipywidgets =8.1.3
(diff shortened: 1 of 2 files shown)


hoechenberger (approved):

recipes/mne-python/construct.yaml
```diff
@@ -69,14 +69,14 @@ specs:
   # - mne-installer-menus =1.4dev0=*_20230503
   # - mne-bids =0.11dev0=*_20221007
   # TODO: ⛔️ ⛔️ ⛔️ DEV BUILDS STOP: CHANGE BEFORE RELEASE! ⛔️ ⛔️ ⛔️
-  - mne =1.7.0=*_1
-  - mne-installer-menus =1.7.0=*_1
+  - mne =1.7.1=*_1
```
hoechenberger, line 72: I suppose we need to reset the build number to 0

PR #273, opened 2024-06-19
ENH: Add mnelab, polars, update to vtk 9.3 and PySide6

### LINKED ISSUE #82: Add MNELAB (requires Qt 6)

### LINKED ISSUE #250: Include Polars
Has started to replace Pandas in new projects:

https://pola.rs

larsoner:
Closes #82
Closes #250

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,7 +154,7 @@ jobs:
         with:
           qt: true
           pyvista: false
-      - run: conda list && conda remove -y qt-main
+      - run: source "${MNE_ACTIVATE}" && conda list && conda remove -y qt-main
       - run: ./tools/check_installation.sh
 
   test_linux:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,7 +154,6 @@ jobs:
         with:
           qt: true
           pyvista: false
-      - run: source "${MNE_ACTIVATE}" && conda list && conda remove -y qt-main
       - run: ./tools/check_installation.sh
 
   test_linux:
@@ -221,6 +220,7 @@ jobs:
           qt: true
           pyvista: false
       - run: ./tools/check_installation.sh
+        timeout-minutes: 5  # < 2 min even on Windows
 
   # Release
   release:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,6 +154,7 @@ jobs:
         with:
           qt: true
           pyvista: false
+      - run: conda list && conda remove -y qt-main
       - run: ./tools/check_installation.sh
 
   test_linux:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-11, macos-14]  # Intel and Apple Silicon, respectively
+        os: [macos-12, macos-14]  # Intel and Apple Silicon, respectively
     runs-on: ${{ matrix.os }}
     env:
       GITHUB_EVENT_NAME: ${{ github.event_name }}
@@ -128,7 +128,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-11, macos-12, macos-13, macos-14]  # macos-14 is Apple Silicon; older versions are Intel
+        os: [macos-12, macos-13, macos-14]  # macos-14 is Apple Silicon; older versions are Intel
     runs-on: ${{ matrix.os }}
     defaults:
       run:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
           path: MNE-Python-*.*
 
   build_linux:
-    name: Build (Linux)
+    name: Build (${{ matrix.os }})
     runs-on: ubuntu-20.04
     defaults:
       run:
@@ -96,7 +96,7 @@ jobs:
           path: MNE-Python-*.*
 
   build_windows:
-    name: Build (Windows)
+    name: Build (${{ matrix.os }})
     strategy:
       fail-fast: false
     runs-on: windows-2019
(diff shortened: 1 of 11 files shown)


hoechenberger:
@larsoner The problem here seems to that `graphviz` pins to `libexpat >=2.6.2`, but `vtk-base` needs an older version (`<2.6`)

hoechenberger:
Actually … no, disregard my previous comment, this doesn't check out at least on my Apple Silicon Mac:

```shell
❯ conda search vtk-base==9.3.0 --info |grep libexpat
  - libexpat >=2.6.2,<3.0a0
  - libexpat >=2.6.2,<3.0a0
  - libexpat >=2.6.2,<3.0a0
  - libexpat >=2.6.2,<3.0a0
  - libexpat >=2.6.2,<3.0a0
❯ conda search graphviz==11.0.0 --info |grep libexpat
  - libexpat >=2.6.2,<3.0a0
```

larsoner:
Now it looks like at least https://github.com/conda-forge/liblsl-feedstock/pull/11 is in the way

larsoner:
Now held up by https://github.com/conda-forge/libtheora-feedstock/issues/21, https://github.com/conda-forge/pyobjc-framework-fsevents-feedstock/pull/39, https://github.com/conda-forge/pyobjc-framework-coreservices-feedstock/pull/20, etc.

larsoner:
macos-11 is no longer available, I'll tweak then we should be good @hoechenberger , feel free to mark for merge-when-green if you're happy!

I dropped yasa because it's blocking a SciPy upgrade and they haven't merged my PR to fix things (and then cut a release), hope that's okay

.github/workflows/build.yml
```diff
@@ -34,10 +34,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: mamba-org/setup-micromamba@v1
         with:
-          condarc: |
-            channels:
-              - conda-forge
-            channel_priority: strict
```
larsoner, line 40: Pretty sure we don't need these based on them saying:

> If nothing is specified, setup-micromamba will create a .condarc next to the micromamba binary with conda-forge as the only channel.

So it should only use conda-forge I think?

hoechenberger: Sounds reasonable to me!

.github/workflows/build.yml
```diff
@@ -72,19 +68,19 @@ jobs:
...
 
   build_linux:
-    name: Build (Linux)
-    runs-on: ubuntu-20.04
+    name: Build (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04]  # Intel and Apple Silicon, respectively
```
hoechenberger, line 76: this comment seems wrong?

larsoner:
Ooof looks like there is a problem with vtk 9.3.0 (and upward) -- it's built against `qt6-main`. One way forward would be to follow https://github.com/conda/constructor/issues/515 and package our own PyQt6. WDYT @hoechenberger ?

larsoner:
... or just hope people don't use plain Python + interactive mode + plotting and stick with PySide6 :(

hoechenberger:
@larsoner More and more conda-forge packages are migrating to qt6 (and hence pyside6). So this is something we'll have to deal with anyway. An alternative would be to switch our installers away from conda and install everything from PyPI. This should work for all platforms except for aarch64. It's what I do in my dev container PR except for vtk and a few other packages (since they're missing for aarch64 on PyPI).

larsoner:
Or maybe someday https://github.com/conda-forge/pyqt-feedstock/pull/136 will land :crossed_fingers:

hoechenberger:
I find this an extremely annoying situation esp since PySide6 is the official Qt binding. Why isn't plotting working properly with it???!

cc @cbrnr

larsoner:
Same as before, https://bugreports.qt.io/browse/PYSIDE-2192

hoechenberger:
yep...

hoechenberger:
@cbrnr Have you had any chance to test whether it's working on Py3.13?

cbrnr:
Not yet, but I will try!

cbrnr:
Unfortunately, PySide6 does not support 3.13 yet.

larsoner:
This is close. One blocker is actually Spyder because it explicitly depends on `pyqt`.

cbrnr:
But why is pyqt problematic?

larsoner:
Because there are only PyQt5 builds (no PyQt6) which means it pulls in qt-main *in addition to* qt6-main and this causes problems

cbrnr:
Ah, I misread that as qtpy. Too many qt py side 5 6 I guess 😄. It's weird that Spyder does not only depend on qtpy though, since I thought they develop and maintain qtpy as an integral part of Spyder...

hoechenberger:
@cbrnr afaik their resources are super limited and they didn't have the time to test with qt6 yet

hoechenberger:
@cbrnr https://github.com/spyder-ide/spyder/issues/20201

hoechenberger:
@larsoner We could consider dropping Spyder and guiding users to the Spyder standalone installers. We'd need to provide some setup script to deploy a proper configuration to make it easy to use MNE from the MNE installers though ...

larsoner:
Hah, apparently we left automerge enabled! But that's okay -- I think for 1.8 we probably should drop Spyder from our installers and point people toward [their standalone installers](https://docs.spyder-ide.org/current/installation.html#standalone-installers). I don't think it'll be too bad to configure it to use our Python distribution/kernel because we install `spyder-kernels` here. But it's something we should do for 1.8 docs. I'll open an issue in MNE-Python and tag for 1.8 so we don't forget.

hoechenberger:
thanks Eric! the inclusion of MNELAB is worth it!

speaking of which, we should probably add it to the installer menu 

it's something we need to take care of in the mne-feedstock repo

larsoner:
Somewhere we have a note about moving the installer menu stuff to this repo, is that still possible? If feel free to give it a shot, or if you know if another project that does it in the constructor I can look at their code and adapt. 

I'll look at Spyder and Napari at least to see if they have something to work from

PR #280, opened 2024-07-02
BUG: Fix bug with artifact download

larsoner:
https://github.com/mne-tools/mne-installers/actions/runs/9764227628/job/26953942126 actually failed so I'll change the merge reqs to have the `Release` step required and sanity check this one by eyeballing the `ls -al ./` step


larsoner:
Okay the `Release` contents `ls -al ./`:

image


Match the ones we'd expect to upload based on the previous release assets:

image

PR #285, opened 2024-07-18
ENH: Make related software list explicit

larsoner:
Draft PR for #12731


larsoner:
Okay in it goes. @hoechenberger if you think stuff should move into or out of this list feel free to in another PR, but hopefully I got the categorization right!

PR #306, opened 2024-11-19
ENH: Add meegkit

larsoner:
Will fail until meegkit actually builds


larsoner:
Now we wait for https://github.com/conda-forge/uv-feedstock/issues/114

PR #313, opened 2024-12-10
TST: Test new version of constructor

larsoner:
One tiny cleanup, and want to verify that `conda-standalone` is indeed the problem, hopefully fails on the first commit then I'll push one to avoid the problematic release :crossed_fingers: 

https://github.com/conda/conda-standalone/issues/121


larsoner:
[Failing](https://github.com/mne-tools/mne-installers/actions/runs/12260643163/job/34206133905?pr=313#step:5:29) as expected on 2a52157. [Fine](https://github.com/mne-tools/mne-installers/actions/runs/12260898323/job/34207012204?pr=313) on b7eedc7.

larsoner:
Okay I suspect the existing `!=` pin is likely correct since the CONDA_QUIET from https://github.com/conda/constructor/pull/950 seems to have fixed things here. So I'll close this and once a new `constructor` is released we can test to make sure it works.

PR #325, opened 2025-01-21
Add wfdb

withmywoessner:
https://github.com/mne-tools/mne-python/pull/13030

### DIFF
--- a/recipes/mne-python/construct.yaml
+++ b/recipes/mne-python/construct.yaml
@@ -194,7 +194,7 @@ specs:
   - py-spy =0.4.0
   - line_profiler =4.1.3
   - memory_profiler =0.61.0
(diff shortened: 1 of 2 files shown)


larsoner (approved):
Looks like we just need to wait for this then restart CIs and merge, marking for merge-when-green thanks in advance @withmywoessner !

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1157114&view=logs&s=ff05ad62-bb9a-53b6-ce9f-72f329a63e7c&j=6d7dd1bc-04f5-5463-bf0b-e9353162f66e

larsoner:
I don't think we have been consistent in adding those, but I guess we could add one here

PR #327, opened 2025-01-28
ENH: Add neo back

larsoner:


larsoner:
Might as well wait on https://github.com/conda-forge/llvmlite-feedstock/pull/96 and https://github.com/conda-forge/numba-feedstock/pull/150

PR #329, opened 2025-02-04
MAINT: Streamline import tests

larsoner:
Rather than list a bunch of `import <mod>` lines, let's parse the YAML file and import systematically. Now imports everything in our related software list, plus a handful of custom imports of other stuff like `PySide6` and `vtk` that it seems like we wanted.

### DIFF
--- a/tests/test_imports.py
+++ b/tests/test_imports.py
@@ -53,7 +53,7 @@ mods = [line[2:].split("#")[0].split(">")[0].split("=")[0].strip() for line in l
 # Plus some custom ones
 mods += """
 darkdetect qdarkstyle numba openpyxl xlrd pingouin questionary
-seaborn plotly pqdm pyvistaqt vtk PySide6.QtCore matplotlib.pyplot
+seaborn plotly pqdm pyvistaqt vtk PySide6 PySide6.QtCore matplotlib matplotlib.pyplot
 """.strip().split()
 if platform.system() == "Darwin":
     mods += ["Foundation"]  # pyobjc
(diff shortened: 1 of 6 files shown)


tests/test_imports.py
```diff
@@ -1,78 +1,81 @@
...
+    "mne-ari",  # https://github.com/john-veillette/mne-ari/pull/7
+}
+for mod in tqdm(mods, desc="Imports", unit="module"):
+    if mod in parsed.ignore:
+        continue
+    use_mod = mod
+    if use_mod.startswith("python-"):  # python-neo
+        use_mod = use_mod[7:]
+    if use_mod.endswith("-py"):  # openneuro-py
+        use_mod = use_mod[:-3]
```
drammock, line 72: if this grows beyond 2 items, I'd switch to using a lookup dict:

```py
mod_name_changes = {"python-neo": "neo", "openneuro-py": "openneuro"}
use_mod = mod_name_changes.get(use_mod, use_mod)
```

tests/test_imports.py
```diff
@@ -1,78 +1,81 @@
...
+    .read_text("utf-8")
+    .splitlines()
+)
+lines = [line.strip() for line in lines]
+all_lines = lines
+sidx = lines.index("# <<< BEGIN RELATED SOFTWARE LIST >>>")
+eidx = lines.index("# <<< END RELATED SOFTWARE LIST >>>")
+lines = [line for line in lines[sidx : eidx + 1] if not line.startswith("#")]
+assert lines
+mods = [line[2:].split("#")[0].split(">")[0].split("=")[0].strip() for line in lines]
```
drammock, line 50:
```suggestion
# NB: next line assumes that there are no "less-than" pins
mods = [line[2:].split("#")[0].split(">")[0].split("=")[0].strip() for line in lines]
```

drammock (approved):
LGTM, 2 nitpicks, feel free to ignore

tests/test_imports.py
larsoner: Yeah it was 3, two of which were `python-` so I figured it was better than a dict

PR #330, opened 2025-02-10
ENH: Add Spyder

### LINKED ISSUE #308: Spyder installation broken on macOS
https://mne.discourse.group/t/mne-18-0-on-macos-m1-wont-launch-spyder/10583

larsoner:
Not sure this will work since locally installing `spyder` from source using `--no-deps` I get an error, but we'll see!

Closes #308


larsoner:
Okay needs https://github.com/conda-forge/spyder-feedstock/pull/230 then maybe it'll work!

larsoner:
Once we merge https://github.com/mne-tools/mne-installers/pull/362 once it's time to come back to this we'll want to add back the Spyder option using something like https://github.com/conda-forge/mne-feedstock/blob/403ca4bf0ea6892cf08cfbffb3b7bd8d251d1b36/recipe/menu/menu.json#L6 plus icons etc.

PR #332, opened 2025-02-11
ENH: Use Python 3.13

larsoner:
Have to remove `sleepecg` but I think everything else (vtk, mayavi, etc.) has been migrated. Might need CDN update though. I'll restart a few times and merge when green.


larsoner:
Going to have to wait for https://github.com/conda-forge/conda-feedstock/pull/249, will remove that from the env temporarily to see if it can then resolve

PR #336, opened 2025-02-28
BUG: Fix bug with Windows post-install script

larsoner:
Currently it fails silently, let's figure out which line is problematic.


assets/post_install_windows.bat
```diff
@@ -13,4 +13,4 @@ echo Pinning BLAS implementation to OpenBLAS.
 echo libblas=*=*openblas >> "%PREFIX%\conda-meta\pinned"
 
 echo Running mne sys_info.
-"%PREFIX%\Scripts\conda" run mne sys_info
+"%PREFIX%\Scripts\conda" run mne sys_info || echo
```
larsoner, line 16: Turns out (I think!) that `mne sys_Info` was the problematic line, so I'll effectively `|| true` it like we do in the macOS script for example

PR #342, opened 2025-04-10
ENH: Add alphacsc

larsoner:
... and document others we can't install.


larsoner:
Needs https://github.com/conda-forge/alphacsc-feedstock/pull/8

PR #362, opened 2025-07-03
ENH: Move menus to this repo

### LINKED ISSUE #360: Move menu stuff to this repo
Scientific Python does it, we should do what they do

https://github.com/scientific-python/installer/tree/main/sp-installer-menu

larsoner:
Closes #360

I think in principle this should work. But I'll test to make sure the console opening works on all three major platforms before merging. Then a follow-up PR can remove `mne-installer-menus` from `mne-feedstock`. I'll probably just roll that into 1.10 release since it's easy.

- [x] Separate menu building as first step (no need to repeat)
- [x] Get working
- [x] Test locally on macOS 
- [x] Test locally on Linux
- [x] Test locally on Windows


hoechenberger:
fantastic initiative, @larsoner 🤩

PR #381, opened 2025-10-01
MAINT: Update intel runner

larsoner:


hoechenberger:
Should we perhaps test Intel on macOS 14 too? This one is not yet deprecated

hoechenberger:
Thanks, @larsoner!

PR #432, opened 2026-04-09
Add welcome and conclusion for Windows, fix Windows JSON

### LINKED ISSUE #431: Check Windows perms
Currently we do
```
icacls %PREFIX% /grant %USERNAME%:(OI)(CI)M /c /t /q
```
in our post install. But looking at the logs nowadays:
```
Setting read/write permissions on the Python installation to runneradmin (this can take a few minutes).
C:\Users\runneradmin\mne-python\1.12.0_0\pkgs\viskores-1.1.0-h509acf7_0\Library\include\viskores-1.1\viskores\filter\scalar_topology\worklet\branch_decomposition\hierarchical_volumetric_branch_decomposer\SuperArcVolumetricComparatorIndirectGlobalIdComparator.h: The system cannot find the path specified.
C:\Users\runneradmin\mne-python\1.12.0_0\pkgs\viskores-1.1.0-h509acf7_0\Library\include\viskores-1.1\viskores\filter\scalar_topology\worklet\contourtree_distributed\boundary_tree_maker\AugmentBoundaryWithNecessaryInteriorSupernodesAppendNecessarySupernodesWorklet.h: The system cannot find the path specified.
C:\Users\runneradmin\mne-python\1.12.0_0\pkgs\viskores-1.1.0-h509acf7_0\Library\include\viskores-1.1\viskores\filter\scalar_topology\worklet\contourtree_distributed\boundary_tree_maker\AugmentBoundaryWithNecessaryInteriorSupernodesUnsetBoundarySupernodesWorklet.h: The system cannot find the path specified.
Successfully processed 506367 files; Failed processing 3 files
Configuring Python to ignore user-installed local packages.
...
Clearing package cache...
Adding C:\Users\runneradmin\mne-python\1.12.0_0\Scripts & Library\bin to PATH...
Setting installation directory permissions...
Successfully processed 256567 files; Failed processing 0 files
```
So we do it in our post-install... then `conda` does something else later.

We should check to see if what `conda` does is already enough. The `icacls` step is slow.

larsoner:
Let's try removing these perms.

Also removing the special-casing of `conda` because I want to check and tinker locally on Windows.

- [x] Add welcome and conclusion for Windows using RTF
- [x] Fix RTF chars
- [x] Check if removing permissions works
- [x] Check and fix NumPy `sys_info` behavior on Windows in MNE-Python
- [x] See if the `conda list --json` works locally or fails, too

Closes #431


larsoner:
On `main` we have no custom Welcome or Conclusion page on Windows, this PR adds them:

image 

image


☝️ These are using the existing RTF file, but it has some characters missing (emojis). I'll push a commit to create a separate `*_windows.rtf` without these characters.

I checked, and permissions setting *is* necessary. Restoring.

My change to make the menus be version `1.0.0` was a mistake, I thought this was arbitrary but it's the number in the Start menu. Worth releasing a new set of installers with this.

PR #440, opened 2026-04-21
BUG: Fix bug with Linux max size

larsoner:
Our release action [failed](https://github.com/mne-tools/mne-installers/actions/runs/24699126101/job/72246241839#step:4:22) to attach the Linux script:
```
Warning: Failed to upload artifact MNE-Python-1.12.1_0-Linux.sh. Validation Failed: {"resource":"ReleaseAsset","code":"custom","field":"size","message":"size must be less than 2147483648"} - https://docs.github.com/rest.
```
I also couldn't upload it manually:

image

So let's:

- [x] Add a check to the installer hash script to `exit 1` if the installer is too big
- [x] Figure out which package is causing the problem by using https://stackoverflow.com/a/67976448/2175965
- [x] Remove package(s) needed to get Linux below 2GB
- [x] Take the resulting artifact and add it to our current release
- [x] Mark PR for merge

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,7 @@ jobs:
           retention-days: 1
 
   build_macos:
+    if: false
     name: Build (${{ matrix.os }})
     needs: [build_local]
     strategy:
@@ -152,6 +153,7 @@ jobs:
           archive: false
 
   build_windows:
+    if: false
     name: Build (${{ matrix.os }})
     needs: [build_local]
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,8 @@ jobs:
   build_local:
     name: Build local packages
     runs-on: ubuntu-latest
+    env:
+      CI_OS: ${{ matrix.os }}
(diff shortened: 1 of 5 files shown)


larsoner:
@drammock looking at the list in

https://github.com/mne-tools/mne-installers/actions/runs/24728521510/job/72337552688?pr=440#step:9:1500

I think we have a few options for what to remove:

1. `r-base` and `rpy2` (allows us to remove a bunch of deps including llvm)
2. `fsleyes` (as it allows us to remove `wxpython` and a bunch of related deps): pypistats "Downloads last month: 1,412"
3. `conda-standalone` and `constructor`

For now I'm going to remove `conda-standalone` and `constructor` since they are much more of a `dev` option, but I wonder about how widely used `fsleyes` actually is, too...

drammock:
r-base and rpy2 should go too IMO, if we can do it without wrecking things.  I was +0.5 on adding the "integration with R" tutorial, but it's been such a can of worms in terms of forcing R installation on our contribs / failing doc builds in the absence of R.

larsoner:
Okay I'll get rid of those here instead. They add a lot of deps !

PR #449, opened 2026-06-09
Add cibuildwheel and abi3audit

larsoner:


larsoner:
Needs https://github.com/conda-forge/cibuildwheel-feedstock/pull/40

PR #456, opened 2026-07-02
Add mne-kit-gui

larsoner:


recipes/mne-python/construct.yaml
```diff
@@ -115,7 +115,7 @@ specs:
   - fsleyes =1.19.0
   - dcm2niix =1.0.20260416
   - dipy =1.12.1
-  - nilearn =0.13.1
+  - nilearn =0.14.0
```
larsoner, line 118: Needs https://github.com/conda-forge/nilearn-feedstock/pull/48

PR #463, opened 2026-08-29
Add pymef

larsoner:
I don't expect it to succeed right away because this needs to hit CDNs


larsoner:
Will need https://github.com/conda-forge/pymef-feedstock/pull/1 and CDNs

